PR #3, opened 2018-01-18
Fix some bugs in smooth transition from Plymouth

besser82:


src/seat-local.c
```diff
@@ -158,6 +158,8 @@ get_vt (SeatLocal *seat, DisplayServer *display_server)
             g_signal_connect (display_server, DISPLAY_SERVER_SIGNAL_READY, G_CALLBACK (display_server_ready_cb), seat);
             g_signal_connect (display_server, DISPLAY_SERVER_SIGNAL_STOPPED, G_CALLBACK (display_server_transition_plymouth_cb), seat);
             plymouth_deactivate ();
+
+            return vt;
```
robert-ancell: Is the problem here that Plymouth is being quit instead of waiting for it to quit? We should check the tests to fix why they aren't detecting this / make them do that. Also I think the else branch below this one should probably also not quite Plymouth.

Can you put this into a separate PR with a better commit description.

src/x-server-local.c
```diff
@@ -481,8 +481,10 @@ x_server_local_start (DisplayServer *display_server)
     if (server->priv->vt >= 0)
         g_string_append_printf (command, " vt%d -novtswitch", server->priv->vt);
 
-    if (server->priv->background)
-        g_string_append_printf (command, " -background %s", server->priv->background);
+    if (!server->priv->background)
+        x_server_local_set_background (server, "none");
+
+    g_string_append_printf (command, " -background %s", server->priv->background);
```
robert-ancell: Right, it seems nothing actually sets the background. I think in Ubuntu we default to none background so I haven't noticed this.

The background shouldn't be set in this method - we should either set the background when we make the object or set the default to "none" (in x_server_local_init).

I think this feature wasn't always available in X servers, but it's probably safe to assume it is. I think there's other features we assume anyway.

Can you put this into a separate PR with a better commit description.

src/seat-local.c
qwertychouskie: Although I agree a theoretically cleaner fix would be nice, given the limited maintenance status of the project, wouldn't it be better to have some form of a fix than none?

robert-ancell:
Closing due to lack of review response. Please re-open if you are able to fix this!

PR #4, opened 2018-01-18
Small bugfixes to Autotools

besser82:


configure.ac
```diff
@@ -107,7 +108,7 @@ if test x"$enable_liblightdm_qt4" != "xno" ; then
     fi
     if test x"$MOC4" = xmoc; then
         if test x"$(readlink $(which $MOC4))" = xqtchooser; then
-            MOC4="$MOC4 --qt=qt4"
+            MOC4="$MOC4 --qt=4"
```
robert-ancell: Is this not working somewhere? I get the following (which suggests 4 and 5 are the recommended value to be passed).

$ qtchooser -l
4
5
default
qt4-x86_64-linux-gnu
qt4
qt5-x86_64-linux-gnu
qt5

configure.ac
```diff
@@ -6,6 +6,7 @@ AC_CONFIG_HEADER(config.h)
 AM_INIT_AUTOMAKE([1.11 no-dist-gzip dist-xz foreign])
 AM_SILENT_RULES(yes)
 LT_INIT
+AC_PROG_CC_STDC
```
robert-ancell: What is occurring with this not set?

Makefile.am
```diff
@@ -1,3 +1,4 @@
+ACLOCAL_AMFLAGS = -I m4
```
robert-ancell: What is occurring with this not set?

besser82: This fixes `libtoolize: Consider adding '-I m4' to ACLOCAL_AMFLAGS in Makefile.am.`.

configure.ac
besser82: Without that being set building lightdm v1.25.1 fails to build on RHEL7 / CentOS7, because the changes in code require the compiler to be run in `-std=c99` mode.

configure.ac
besser82: This was a downstream patch we carried in Fedora / EPEL since some long time.  The patch actually changes to an invocation with the recommend values, as you described.

PR #5, opened 2018-01-18
Add an option to disable writing dmrc files

besser82:
This might be needed for some distributions, which are using SELinux.  See: https://bugzilla.redhat.com/show_bug.cgi?id=963238

### DIFF
--- a/common/dmrc.c
+++ b/common/dmrc.c
@@ -13,6 +13,7 @@
 #include <string.h>
 #include <unistd.h>
 
+#include "config.h"
 #include "dmrc.h"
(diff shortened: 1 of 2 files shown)


robert-ancell:
I think a compile option is the wrong way to solve this - it should either be disabled in configuration or not done if AccountsService is present.

I think I kept supporting both in case you switched between them, but it probably reasonable to ignore .dmrc if AccountsService is present.

robert-ancell:
Closing due to lack of review response. Please re-open if you are able to fix this!

PR #13, opened 2018-03-23
in "Remove X authority": Ignore any error & don't exit, continue closing the session

imz:
...to close PAM etc. anyway.

Otherwise the PAM session might be left open, resources not unmounted/freed, etc.

### DIFF
--- a/src/session-child.c
+++ b/src/session-child.c
@@ -771,8 +771,7 @@ session_child_run (int argc, char **argv)
 
         if (error)
             g_printerr ("Error removing X authority: %s\n", error->message);
-        if (!result)
-            _exit (EXIT_FAILURE);
+        /* Ignore this error, don't exit, continue closing the session. */
     }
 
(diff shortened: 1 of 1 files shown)


robert-ancell:
Good catch!

imz:
Thanks for review and the merge!

Now I've just got another issue in my mind:

in earlier versions, there used to be a `g_clear_error (&error);` at [this place](https://github.com/CanonicalLtd/lightdm/pull/12/files#diff-8f219dfe989f537dd3f4f6e579c6a4cdL807), and this was not quite important if `_exit` was done afterwards. But now, we are not doing `_exit` anymore, perhaps, it might make sense to put `g_clear_error (&error);` back?

What do you think?

robert-ancell:
Right - it should really clear the error. Note in this case it doesn't matter because we don't use 'error' again and soon exit anyway.

PR #26, opened 2018-06-21
Remove Unity System Compositor support - it is a dead project

robert-ancell:


robert-ancell:
See https://github.com/CanonicalLtd/lightdm/issues/24

PR #28, opened 2018-06-25
document the XDG_SEAT_PATH environment variable

tpo:
This fixes https://github.com/CanonicalLtd/lightdm/issues/27


follower:
As far as I can tell, the correct path seems to contain "DisplayManager" rather than "DesktopManager"?

Could you confirm this?

(https://github.com/canonical/lightdm/pull/158)

PR #78, opened 2019-06-07
Fix environment cleaning without clearenv()

madpilot78:
On systems without `clearenv()` the `environ` variable must be imported as external before it can be manipulated to reset the environment.

This is, for example, necessary on FreeBSD, otherwise the effect of `environ = NULL` is a compiler error or in some cases to create a new local variable.

### DIFF
--- a/src/process.c
+++ b/src/process.c
@@ -73,6 +73,10 @@ static GHashTable *processes = NULL;
 static pid_t signal_pid;
 static int signal_pipe[2];
 
+#ifndef HAVE_CLEARENV
+extern char **environ;
+#endif
+
 Process *
 process_get_current (void)
 {


robert-ancell:
Thanks @madpilot78 !

triaxx:
On NetBSD, `environ = NULL` followed by `setenv()` produces a core dumped. Since `environ` is a variable of type `char**`, I think it is a NULL-terminated array of `char*` where each entry corresponds to a environment variable. The following code:
```
setenv("TEST", "test", 1);
printf("TEST=%s\n", getenv("TEST"));
*environ = NULL;
printf("TEST=%s\n", getenv("TEST"));
```
produces the following output:
```
TEST=test
TEST=(null)
```
Do you think `*environ = NULL` could be a portable method to fix systems that do not implement `clearenv()`

madpilot78:
Not sure.

When I proposed this patch I did some research and found no description of portable ways to handle this.

I could try a test in FreeBSD and see how things work out, but I need a little time for that.

PR #85, opened 2019-07-30
Move D-Bus conf file to $(datadir)/dbus-1/system.d

Soapux:
Since D-Bus 1.9.18 configuration files installed by third-party should
go in $(datadir)/dbus-1/system.d.  The old location is for sysadmin
overrides.

https://lists.freedesktop.org/archives/dbus/2015-July/016746.html


robert-ancell:
Thanks @Soapux!

PR #104, opened 2019-12-10
Determine a session is wayland with a better heuristic

worldofpeace:
Using the compiled in string WAYLAND_SESSIONS_DIR had some problems.
In environments like NixOS, which install packages under their
own immutable prefix under /nix/store. this assumption that the package
would be compiled with `datadir=/usr/share` will break down. Checking
if the directory contains the appropriate string is much simpler and will
work with platforms like NixOS (or others as this setup popularizes).

### DIFF
--- a/liblightdm-gobject/session.c
+++ b/liblightdm-gobject/session.c
@@ -171,7 +171,7 @@ load_sessions (const gchar *sessions_dir)
     {
         const gchar *default_type = "x";
 
-        if (strcmp (dirs[i], WAYLAND_SESSIONS_DIR) == 0)
+        if (dirs[i] != NULL && g_str_has_suffix (dirs[i], "/wayland-sessions") == TRUE)
             default_type = "wayland";
 
         sessions = load_sessions_dir (sessions, dirs[i], default_type);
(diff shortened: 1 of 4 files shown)


liblightdm-gobject/session.c
```diff
@@ -171,7 +171,7 @@ load_sessions (const gchar *sessions_dir)
     {
         const gchar *default_type = "x";
 
-        if (strcmp (dirs[i], WAYLAND_SESSIONS_DIR) == 0)
+        if (dirs[i] != NULL && g_str_has_suffix(dirs[i], "/wayland-sessions")  == TRUE)
```
robert-ancell: Missing whitespace, i.e. `has_suffix (`

src/seat.c
```diff
@@ -956,7 +956,7 @@ find_session_config (Seat *seat, const gchar *sessions_dir, const gchar *session
     for (int i = 0; dirs[i]; i++)
     {
         const gchar *default_session_type = "x";
-        if (strcmp (dirs[i], WAYLAND_SESSIONS_DIR) == 0)
+        if (dirs[i] != NULL && g_str_has_suffix(dirs[i], "/wayland-sessions")  == TRUE)
```
robert-ancell: Missing whitespace, i.e. `has_suffix (`

worldofpeace:
@robert-ancell I've fixed the whitespace. Thanks for the prompt response :sparkles:

robert-ancell:
Thanks @worldofpeace!

PR #111, opened 2020-02-07
Only set LANGUAGE if not in a live session

gunnarhj:
Fixes LP: #1861481


gunnarhj:
This change was uploaded to Ubuntu/focal as `1.30.0-0ubuntu3.1`.

Issue #18, opened 2018-05-14
Support non-root X 


leigh123linux:
Currently lightdm runs Xorg as root

```
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root       851  0.9  1.5 751600 254880 tty1    Ssl+ May11  43:46 /usr/libexec/Xorg -core -noreset :0 -seat seat0 -auth /var/run/lightdm/root/:0 -nolisten tcp vt1 -novtswitch -background none
```

https://bugs.launchpad.net/lightdm/+bug/1292324

Acatorn:
Hello there,
First of all, many thanks to everyone commited to LightDM. It is really a great and much needed alternative to GDM and SDDM. Any news on that issue? Nowadays it is pretty common to run X as normal user but in order to do that one must use xinit or GDM. It would be great if LightDM would support such important (from security point of view) feature. @robert-ancell is implementing rootless LightDM is possible in any near future?

montvid:
What's the problem with fixing this in lightdm? Just uninstalled lightdm from Debian 11 and got rootless X on xfce. Today it is insane to run xorg as root!?

lproven:
Bumping this, since there's a new release. :-)

montvid:
It is beyond me why all the distro's I tested run X as root? Just because of lightdm? 😂

doubledge-spec:
Any updates from devs? Would really like to run rootless X.org.

eworm-de:
Just another [xorg security release](https://lists.x.org/archives/xorg/2023-February/061255.html) for an issue that can be exploited when running privileged. Any chance we see some progress here?

DanielGibson:
Xubuntu 24.04 still running Xorg as root, presumably because it uses lightdm, is not cool.

zaryc89:
forgive me if I have this wrong, rootless can be good, but, rootless is not a one-line solution. userspace and namespaces have their own security issues. mitigatng the xorg keystrokes is understandably something we want fixed.

hamishmb:
Is this still something that lightDM does? Linux Mint, for example, still uses lightDM so it's worth knowing.

hamishmb:
Confirmed with `ps -ef | grep X` that Xorg is still running as root.

corsac-s:
Hey,

[...]

PR #127, opened 2020-06-10
Add ability to drop privileges when launching X server

twhitehead:
We recently experienced a local root escalation issue, and our security team identified one possible avenue for the issue having been one of these security bugs in the tigervnc server along with the fact that it was running as root despite just being a glorified in memory frame buffer

https://access.redhat.com/errata/RHSA-2020:1497

We've patched it, but, to protect against any further such attacks in the future, I also went ahead and added the ability for lightdm to drop privileges when forking local X servers

```
$ ps -HA -o pid,user,group,args
...
3896 root     root       /usr/sbin/lightdm
 3906 vnc      vnc          /usr/bin/Xvnc -securitytypes vencrypt ...
 3925 root     root         lightdm --session-child 19 22
 3928 lightdm  lightdm        /usr/sbin/slick-greeter
 3965 root     root         lightdm --session-child 15 22
...
```

The option is exposed as a `[VNCServer]` `user` setting, it's not VNC specific. The mechanism was added entirely to the x-server-local and process infrastructures, so it would be pretty trivial to apply it to the X sever proper too (which I've heard can now run without root in some cases too).

I did have to rebase my original commits (which I tested) due to a recent commit that removed the cached priv pointers x-server-local and process, changing many lines of `sever->priv->...` to just `priv->...`. It was a pretty trivial fix, so I haven't verified it beyond that fact that it compiles.


twhitehead:
Just a note that it doesn't look like that CI build test failure was an issue with this pull request. The Fedora build was fine. The Ubuntu build seems to have hung configuring the build environment without ever actually reaching the build step.

src/x-server-local.c
```diff
@@ -391,6 +394,44 @@ got_signal_cb (Process *process, int signum, XServerLocal *server)
...
+static gboolean
+poll_for_socket_cb (XServerLocal *server)
+{
+    XServerLocalPrivate *priv = x_server_local_get_instance_private (server);
+
+    /* Check is X11 socket file exists as an alternative startup test to SIGUSR1 */
+    GStatBuf statbuf;
+    g_autofree gchar *socketpath = g_strdup_printf ("/tmp/.X11-unix/X%d", priv->display_number);
+
+    if ( g_stat (socketpath, &statbuf) == 0 )
```
robert-ancell, line 406: Extra whitespace here doesn't match existing code.

src/x-server-local.c
```diff
@@ -391,6 +394,44 @@ got_signal_cb (Process *process, int signum, XServerLocal *server)
...
+        else {
+            priv->got_signal = TRUE;
+            l_debug (server, "Detected valid X11 socket for X server :%d", priv->display_number);
+
+            // FIXME: Check return value
+            DISPLAY_SERVER_CLASS (x_server_local_parent_class)->start (DISPLAY_SERVER (server));
+        }
+
+        /* Cancel the timer as the file showed up */
+        return FALSE;
```
robert-ancell, line 428: Change the TRUE and FALSE to G_SOURCE_REMOVE/G_SOURCE_CONTINUE.

src/x-server-local.c
```diff
@@ -391,6 +394,44 @@ got_signal_cb (Process *process, int signum, XServerLocal *server)
...
+    /* Check is X11 socket file exists as an alternative startup test to SIGUSR1 */
+    GStatBuf statbuf;
+    g_autofree gchar *socketpath = g_strdup_printf ("/tmp/.X11-unix/X%d", priv->display_number);
+
+    if ( g_stat (socketpath, &statbuf) == 0 )
+    {
+        uid_t uid = priv->user ? user_get_uid(priv->user) : 0;
+
+        /* It has to be a valid socket file */
+        if (!(statbuf.st_mode & S_IFSOCK))
```
robert-ancell, line 411: The if statement is a little hard to follow here - I think it would be simpler if you just return from the function, i.e.

```
if (!(statbuf.st_mode & S_IFSOCK))
   return G_SOURCE_REMOVE;
```

corsac-s:
Hi, as a Debian maintainer I'll try to test the patch and report back, not for the XVnc part but rather for the whole “deprivilege ” thing. We have https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=809067 opened which I fowarded to https://bugs.launchpad.net/lightdm/+bug/1292324 on Launchpad but that bug was never migrated to github I think (although there is #18 )

src/x-server-local.c
twhitehead: I believe you mean blank like 405. I've deleted it. Please let me know if this wasn't it.

src/x-server-local.c
twhitehead: Thanks. I've made the change.

src/x-server-local.c
twhitehead: I agree it is a bit hard to read. I've changed it as you suggested.

twhitehead:
@corsac-s Although most of the changes were made at the `process.c` and `xserver-local.c` levels, currently only the `seat-xvnc.c` code has been changed to call the new `x_server_local_set_user` function when the `[VNCServer]` `user` option is set that triggers the privilege drop.

A similar configuration option and `x_server_local_set_user` call would have to be made for the case of non-xvnc local seat to enable rootless regular X. From the issues you link though, it does sound like there may also be some VT ownership complications that would need be resolved (it isn't clear to me how exactly the Xorg VT stuff works).

twhitehead:
@robert-ancell I think it is reasonable if this requires a newer Xorg. I'll see what I can do with regard to the `-display-fd` method of getting the display.

twhitehead:
Just a note that the `Xserver` man page says that `-displayfd` means that the X server chooses a display number for you (not sure what that might mean for lightdm).

> -displayfd fd
>
> specifies  a  file  descriptor in the launching process.  Rather than specify a display number, the X server will attempt to listen on successively higher display numbers, and upon finding a free one, will write  the  display  numberback on this file descriptor as a newline-terminated string.  The -pn option is ignored when using -displayfd.

src/x-server-local.c
robert-ancell: I mean it should change from `if ( g_stat (socketpath, &statbuf) == 0 )` to `if (g_stat (socketpath, &statbuf) == 0)`

robert-ancell:
> Just a note that the `Xserver` man page says that `-displayfd` means that the X server chooses a display number for you (not sure what that might mean for lightdm).
> 
> > -displayfd fd
> > specifies  a  file  descriptor in the launching process.  Rather than specify a display number, the X server will attempt to listen on successively higher display numbers, and upon finding a free one, will write  the  display  numberback on this file descriptor as a newline-terminated string.  The -pn option is ignored when using -displayfd.

Looking at the xorg source:

```c
       else if (argv[i][0] == ':') {
            /* initialize display */
            display = argv[i];
            explicit_display = TRUE;
```

```c
    else if ((displayfd < 0) || explicit_display) {
        if (TryCreateSocket(atoi(display), &partial) &&
            ListenTransCount >= 1)
            if (!PartialNetwork && partial)
                FatalError ("Failed to establish all listening sockets");
    }
    else { /* -displayfd and no explicit display number */
        Bool found = 0;
        for (i = 0; i < 65536 - X_TCP_PORT; i++) {
            if (TryCreateSocket(i, &partial) && !partial) {
                found = 1;
                break;
            }
```

If you pass a display number (as we do in LightDM) it won't try to pick it's own. I'm pretty sure I got this to work in the past so it should work fine.

twhitehead:
Thanks for looking into that Robert. I will work to incorporate this as the display-server-has-started check instead of polling for the socket file.

Just a heads up that it may take awhile as I am now heavily involved in another project, but I will push the results here when done.  :smile:

twhitehead:
@corsac-s I spent a bit of time reading up on how the whole seat thing and such is all done. The best resource I found was this

https://dvdhrm.wordpress.com/2013/08/24/session-management-on-linux/

and combined with the fact that I believe I saw session dbus calls in the lightdm code, I think there is a reasonable chance that my device ownership and such concerns are null and void.

corsac-s:
> @corsac-s I spent a bit of time reading up on how the whole seat thing and such is all done. The best resource I found was this
> 
> https://dvdhrm.wordpress.com/2013/08/24/session-management-on-linux/
> 
> and combined with the fact that I believe I saw session dbus calls in the lightdm code, I think there is a reasonable chance that my device ownership and such concerns are null and void.

So that would only leave the specific option and the  `x_server_local_set_user` stuff?

montvid:
When is this shipping? :( I had to uninstall lightdm to make xorg run as nonroot.

eylles:
is it possible to use the `xserver-command` option form the config along xserver-xorg-legacy to force x11 to drop privileges even when started by lightdm or trying that is just an exercise in futility?

twhitehead:
> When is this shipping? :( I had to uninstall lightdm to make xorg run as nonroot.

The developers wanted a better detection for when X11 is ready for connections than socket polling (see above comment about `-displayfd`) and I haven't got around to implementing that yet.

Hopefully I'll have some time at some point in the future here. It likely won't be in the nearish future, though, so, if anyone else is feeling up for it, go for it!

zWhdmB5T:
@twhitehead Hi! Is there still any effort on this?

twhitehead:
@zWhdmB5T it is something I would like to do sometime, not sure if it will ever get up high enough on my priority list to get done though...

PR #195, opened 2021-05-24
Do not print errors when pam_kwallet is not installed

oliwer:
This fixes Launchpad bug [#1309535](https://bugs.launchpad.net/ubuntu/+source/lightdm/+bug/1309535/)

Since kwallet is an optional dependency concerning only KDE users,
tell PAM to not print error messages if the kwallet modules are
not present. This is done simply by prefixing the corresponding
lines with a '-' (dash), as suggested in pam.conf(5).


oliwer:
@robert-ancell Any comment?

oliwer:
Apparently this is being fixed in #216 :unamused:

PR #196, opened 2021-06-16
fix: use g_autofree in x_authority_load

KT-lcz:
add g_autofree before guint8 *xauth_data;


robert-ancell:
Thanks @KT-lcz!

KT-lcz:
> Thanks @KT-lcz!
use one commit to resolve this.
https://github.com/canonical/lightdm/pull/197

PR #223, opened 2022-01-12
Fix: Don't disconnect signals on cancel before the session actually ends

wolneykien:
Otherwise, no completion message is sent to the greeter leaving it in the `in_authentication` state.

### DIFF
--- a/src/greeter.c
+++ b/src/greeter.c
@@ -63,6 +63,9 @@ typedef struct
     /* API version the client can speak */
     guint32 api_version;
 
+    /* TRUE if the PAM session is being cancelled */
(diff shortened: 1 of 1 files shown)


wolneykien:
@robert-ancell , don't you know how can I reproduce the failing tests? The branch is building normally and all tests are passed — at least, in ununtu:rolling Docker container… The same thing with the patch rebased on `main`.

robert-ancell:
@wolneykien are you planning on rebasing this?

wolneykien:
Yes, Robert. I forgot about the issue. Let me done that, say, in a couple of days..?

PR #226, opened 2022-01-17
fix unlocalized greeter messages

liaohanqin:
pam message with error language env.


robert-ancell:
What problem is this PR solving? Should the PAM conversation not inherit the system locale?

lo1ol:
> What problem is this PR solving? Should the PAM conversation not inherit the system locale?

Pam modules not inherit system locale. Pam application should set it explicitly. Actually applications don't use system locale on startup. If you don't call `setlocale(LC_ALL, "")` [the default locale will be "C"](https://man7.org/linux/man-pages/man3/setlocale.3.html):

>  On startup of the main program, the portable "C" locale is
> selected as default.  A program may be made portable to all
> locales by calling:
> setlocale(LC_ALL, "");

System and parent's process locale don't matter
```bash
cat > main.c << EOF
#include <stdio.h>
#include <libintl.h>
#include <locale.h>
#include <stdlib.h>

int main() {
   printf("%s\n", dgettext("Linux-PAM", "Password: "));
   setlocale(LC_ALL, "");
   printf("%s\n", dgettext("Linux-PAM", "Password: "));
   system("./main2");
}
EOF

cat > main2.c << EOF
#include <stdio.h>
#include <libintl.h>
#include <locale.h>

int main() {
   printf("%s\n", dgettext("Linux-PAM", "Password: "));
}
EOF

gcc main.c -o main
gcc main2.c -o main2

LANG=ru_RU.UTF-8 ./main
```

My output is:
```
Password: 
Пароль: 
Password: 
```

So, this change really helps.

I think PR's author abandoned it. If you have no time to support it, I could

jpeisach:
Actually, testing using `LC_ALL=es_ES.utf8 lightdm --test-mode --debug` works. Maybe it's untranslated locales?

lo1ol:
> Actually, testing using LC_ALL=es_ES.utf8 lightdm --test-mode --debug works. Maybe it's untranslated locales?

Are you talking about messages from pam modules? Or about text in greeter? This PR fixes locale messages from PAM modules

jpeisach:
> Are you talking about messages from pam modules? Or about text in greeter? This PR fixes locale messages from PAM modules

I was thinking about the text in the greeter.. is there something else I am missing?

I didn't think lightdm was responsible for setting system locale

fthaltun:
@jpeisach @robert-ancell 

LightDM already passes the locale environment variables to PAM, but the session-child process itself still starts with the C locale because setlocale() has not been called.

PAM modules such as pam_sss use gettext for localized messages. As a result, they see the C locale and fall back to English, even though LANG and LC_* are present in the environment.

Calling setlocale (LC_ALL, "") before pam_start() initializes the process locale from the existing environment without changing the system locale. This allows PAM messages to use the expected locale.

The screenshots below show the difference with and without this patch:

Without this patch: the PAM message is displayed in English.
without-this-patch

With this patch: the same PAM message is correctly displayed in the configured locale.
with-this-patch

PR #233, opened 2022-02-08
Don't call setenv with a NULL value - the behaviour is undefined.

robert-ancell:


qwertychouskie:
Was there any particular reason this one wasn't merged?  The CI seems to have bugged out but the code itself looks good.

Issue #165, opened 2021-01-29
Not showing any login screen 


Chaitanyabsprip:
There is no screen showing on boot. Only an underscore blinking in the top right of the screen. I can access the tty and login from there but light dm is not working. 
```
-$ lightdm --show-config                                                                                 
   [LightDM]
A  logind-check-graphical=true
A  run-directory=/run/lightdm

   [Seat:*]
A  greeter-session=lightdm-gtk-greeter
A  user-session=qtile
A  session-wrapper=/etc/lightdm/Xsession
A  greeter-setup-script=/usr/bin/numlockx on

Sources:
A  /etc/lightdm/lightdm.conf
```

BeNeDeLuX:
Confirm! I have the same problem under Manjaro. I have also to switch to tty2 and "startxfce4" in my case.

Here is my LightDM Config:
```
lightdm --show-config
   [LightDM]
A  run-directory=/run/lightdm

   [Seat:*]
A  session-wrapper=/etc/lightdm/Xsession
A  autologin-user=benedelux
A  autologin-user-timeout=0

Sources:
A  /etc/lightdm/lightdm.conf
```
Here are all Informations about my System setup:

```
System:    Kernel: 5.11.1-1-MANJARO x86_64 bits: 64 compiler: gcc v: 10.2.1 Desktop: Xfce 4.16.0 Distro: Manjaro Linux 
Machine:   Type: Laptop System: TUXEDO product: TUXEDO Pulse 14 Gen1 v: Standard serial: <filter> 
           Mobo: TUXEDO s model: PULSE1401 v: Standard serial: <filter> UEFI: American Megatrends v: N.1.07.A02 
           date: 12/08/2020 
Battery:   ID-1: BAT0 charge: 29.0 Wh condition: 46.7/46.7 Wh (100%) model: standard status: Discharging 
CPU:       Info: 8-Core model: AMD Ryzen 7 4800H with Radeon Graphics bits: 64 type: MT MCP arch: Zen 2 rev: 1 L2 cache: 4 MiB 
           flags: avx avx2 lm nx pae sse sse2 sse3 sse4_1 sse4_2 sse4a ssse3 svm bogomips: 92666 
           Speed: 2224 MHz min/max: 1400/2900 MHz boost: enabled Core speeds (MHz): 1: 2224 2: 2076 3: 1689 4: 2117 5: 1645 
[...]

PR #237, opened 2022-03-09
Avoid race condition between graphics driver and lightdm

gsilvan:
There is a possible race condition where lightdm fails to a black screen/tty, if lightdm loads faster than the system graphics driver. There is already the setting `logind-check-graphical` but it is unset per default.
As more and more people run into this pitfail, this PR sets `logind-check-graphical=True`.

**References**
https://wiki.archlinux.org/title/LightDM#LightDM_does_not_appear_or_monitor_only_displays_TTY_output
https://www.reddit.com/r/linuxmasterrace/comments/p9ebi0/logindcheckgraphicaltrue/
https://askubuntu.com/questions/1360285/login-screen-issue-first-lightdm-needs-to-be-killed-to-get-x-session-to-work
https://github.com/canonical/lightdm/issues/165

### DIFF
--- a/data/lightdm.conf
+++ b/data/lightdm.conf
@@ -26,7 +26,7 @@
 #lock-memory=true
 #user-authority-in-system-dir=false
 #guest-account-script=guest-account
-#logind-check-graphical=false
+#logind-check-graphical=true
 #log-directory=/var/log/lightdm
 #run-directory=/var/run/lightdm
 #cache-directory=/var/cache/lightdm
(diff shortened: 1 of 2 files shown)


gsilvan:
signed

qwertychouskie:
CI needs re-trigger but the change itself checks out.

robert-ancell:
Looks like in 2014 this [feature was unreliable](https://github.com/canonical/lightdm/commit/9c8428577432433a3fd9c033ba68f42595ea42dd). I'd hope that is no longer the case!

robert-ancell:
Tests passed locally, so this should be good to land, thanks @gsilvan!

tangruize:
It seems that `logind-check-graphical=true` causes LightDM in LXC container failed to start. I created an issue https://github.com/canonical/lightdm/issues/263 .

PR #238, opened 2022-03-10
Set layout with xkl_config and default_layout

### LINKED ISSUE #225: Inconsistent layout management
## The problem

LightDM does not change the `default_layout` when using `lightdm_set_layout`, which provokes an inconsistency between the `lightdm_get_layout` and `lightdm_set_layout` functions. In other words, getting the layout does not return the set layout.

Exactly, here is the issue:

https://github.com/canonical/lightdm/blob/a02b4ee578f99a9c34fc98f2a4a4cde2f2a25053/liblightdm-gobject/layout.c#L158-L168

The layout is set in a new `XklConfigRec`, which works, but doesn't change the default `XklConfigRec *xkl_config`.

### Use the default `xkl_engine`

Instead of creating a new `XklConfigRec`, use the existing one: `xkl_config`. To prevent of `xkl_config` being NULL, call `lightdm_get_layouts()` as it does in `lightdm_get_layout` to initialize `xkl_config`.

Maybe something like this?: (not tested)

```c
void
lightdm_set_layout (LightDMLayout *dmlayout)
{
    g_return_if_fail (dmlayout != NULL);
    lightdm_get_layouts ();

    // get the `layout` and `variant` from `dmlayout`

    config->layouts[0] = g_steal_pointer (&layout);
    config->variants[0] = g_steal_pointer (&variant);

    if (!xkl_config_rec_activate (xkl_config, xkl_engine))
        g_warning ("Failed to activate XKL config");
}
```

### Set `default_layout`

Keep the current `lightdm_set_layout` method, but set `default_layout` as `dmlayout`: (not tested)

```c
void
lightdm_set_layout (LightDMLayout *dmlayout)
{
    // check if dmlayout
    // get the `layout` and `variant` from `dmlayout`
    // Do the setting process

    default_layout = dmlayout;
}
```

JezerM:
Fixes #225

## Changes

- Use global `xkl_config` instead of creating a new one each time `lightdm_set_layout` is called.
- Set `default_layout` as `dmlayout`, allowing to get the new layout with `lightdm_get_layout`.

## Behavior before and now

> Examples made with node-gtk for Node.js

Before, `lightdm_get_layout` always returned the first catched layout:

```sh
❯ node
Welcome to Node.js v16.14.0.
> const gi = require("node-gtk");
> const LightDM = gi.require("LightDM", "1");
> LightDM.getLayout().name;
'latam'
> let layout = LightDM.getLayouts().find((v) => v.name == "us");
> LightDM.setLayout(layout);
> LightDM.getLayout().name;
'latam'
```

Now, `lightdm_get_layout` returns the new layout:

```sh
❯ node
Welcome to Node.js v16.14.0.
> const gi = require("node-gtk");
> const LightDM = gi.require("LightDM", "1");
> LightDM.getLayout().name;
'latam'
> let layout = LightDM.getLayouts().find((v) => v.name == "us");
> LightDM.setLayout(layout);
> LightDM.getLayout().name;
'us'
```

### DIFF
--- a/liblightdm-gobject/layout.c
+++ b/liblightdm-gobject/layout.c
@@ -148,6 +148,7 @@ void
 lightdm_set_layout (LightDMLayout *dmlayout)
 {
     g_return_if_fail (dmlayout != NULL);
+    lightdm_get_layouts();
 
     g_debug ("Setting keyboard layout to '%s'", lightdm_layout_get_name (dmlayout));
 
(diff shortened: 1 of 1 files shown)


JezerM:
So, it seems the action cannot detect the signed CLA, though it's signed...

JezerM:
CLA signed, finally :D

robert-ancell:
Could you move some of the PR description into the commit message? I'm just worried the reasoning for this change might get lost in the future.

JezerM:
Done~

PR #244, opened 2022-04-19
fix: block login if shell in nologin or false

KT-lcz:
When the account's shell is /usr/sbin/nologin or /usr/bin/false,disable autologin


KT-lcz:
I have signed contributor agreement just a moment ago.Thanks for reminding.
@robert-ancell

PR #245, opened 2022-04-27
Use the first available session config as default

wolneykien:
If there is no "default.desktop" session then use the first available .desktop file as default.


robert-ancell:
The concern about this is:
- A configuration failure can lead to a session being chosen that was not configured to be allowed. This might break a system which has intentionally limited which sessions are available.
- This is a behaviour change, which might break an existing install. To avoid this a configuration option is usally used to opt-in to this new behaviour.

PR #246, opened 2022-04-27
Add the `login-unknown` configuration option

wolneykien:
If the option is set to `True`, LightDM proceeds to start the new session even if the username is not known. Intended to be used with `pam_mkuser.so` PAM module.

### DIFF
--- a/src/seat.c
+++ b/src/seat.c
@@ -998,8 +998,9 @@ create_user_session (Seat *seat, const gchar *username, gboolean autostart)
     l_debug (seat, "Creating user session");
 
     /* Load user preferences */
+    gboolean login_unknown = config_get_boolean (config_get_instance (), "LightDM", "login-unknown");
     g_autoptr(User) user = accounts_get_user_by_name (username);
-    if (!user)
+    if (!user && !login_unknown)
     {
         l_debug (seat, "Can't login unknown user '%s'", username);
         return NULL;
--- a/src/seat.c
+++ b/src/seat.c
@@ -998,7 +998,13 @@ create_user_session (Seat *seat, const gchar *username, gboolean autostart)
     l_debug (seat, "Creating user session");
 
     /* Load user preferences */
-    gboolean login_unknown = config_get_boolean (config_get_instance (), "LightDM", "login-unknown");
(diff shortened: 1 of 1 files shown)


src/seat.c
```diff
@@ -998,8 +998,15 @@ create_user_session (Seat *seat, const gchar *username, gboolean autostart)
     l_debug (seat, "Creating user session");
 
     /* Load user preferences */
+    gboolean login_unknown = FALSE;
+
+    if (seat_get_string_property (seat, "login-unknown"))
```
robert-ancell, line 1003: Please add this into `data/lightdm.conf` with the default value and some documentation.

src/seat.c
```diff
@@ -998,8 +998,15 @@ create_user_session (Seat *seat, const gchar *username, gboolean autostart)
     l_debug (seat, "Creating user session");
 
     /* Load user preferences */
+    gboolean login_unknown = FALSE;
+
```
robert-ancell, line 1002: Nit: remove empty line.

src/seat.c
```diff
@@ -998,8 +998,15 @@ create_user_session (Seat *seat, const gchar *username, gboolean autostart)
...
+    if (seat_get_string_property (seat, "login-unknown"))
+        login_unknown = seat_get_boolean_property (seat, "login-unknown");
```
robert-ancell, line 1004: login-unknown is a bit of a cryptic name, I suggest "allow-empty-username"

robert-ancell:
This should also have a test, see `tests/scripts/login-long-username.conf` for one to copy. (I appreciate the tests are a bit hard to write, so please have a go but if not I'll add them when I get time).

robert-ancell:
Actually should really add two tests (empty username, one with config option and one without), as we weren't testing an empty username.

src/seat.c
robert-ancell: Now I look at this again, it is unknown users, not just empty usernames.

robert-ancell:
Now I look at this a second time, I think the correct solution is to always allow this. I can't see a reason for us to fail at this point - it should fail in the PAM layer if the username is incorrect.

PR #280, opened 2022-11-11
dm-tool: Drop mentions of `add-seat`

### LINKED ISSUE #279: dm-tool: `add-seat` poorly documented
Hi. [This](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=849072) was reported on Debian BTS.

The code and the error message suggest that the `AddSeat()` method is deprecated. What's more is that **dm-tool**(1) doesn't explain what arguments are expected when running `dm-tool add-seat`.

So, there seems to be two closely related issues:

1. `add-seat TYPE` isn't explained at all. It's not obvious what `TYPE` refers to here.
2. Its deprecation isn't documented in the man page.

Considering that https://github.com/canonical/lightdm/blob/de09f2c34175b44e12c4a49c87905812576f9661/src/display-manager-service.c#L229
is always called whenever `AddSeat()` is invoked, I think it'd be best to mention its deprecation in the man page and drop `add-seat` from Bash completion.

imageck:
This command is deprecated and isn't well-documented anyway. Close #279.

### DIFF
--- a/data/bash-completion/dm-tool
+++ b/data/bash-completion/dm-tool
@@ -4,7 +4,7 @@ _dm_tool()
 {
     local cur prev opts
     _init_completion || return
-    opts='switch-to-greeter switch-to-user switch-to-guest lock list-seats add-nested-seat add-local-x-seat add-seat'
+    opts='switch-to-greeter switch-to-user switch-to-guest lock list-seats add-nested-seat add-local-x-seat'
 
     case "$prev" in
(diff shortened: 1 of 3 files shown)


imageck:
LightDM isn't a Canonical project anymore, is it?

Anyway, I'm not signing any CLA. Such a shame.

robert-ancell:
Closing as contributor doesn't want to sign the CLA.

Issue #95, opened 2019-10-15
Make location of .xsession-errors configurable


h88e22dgpeps56sg:
LightDM always uses `$HOME/.xsession-errors` as the log file because it is harcoded. There are several possible reasons for someone to define another location, for example to write it into the RAM or to just have a tidier `$HOME`-folder.

`$HOME` just isn't the right place for the file. `$XDG_RUNTIME_DIR` (that's `/run/user/$UID` with systemd) or `$XDG_CACHE_HOME` (= `$HOME/.cache`) would be more suitable.

Currently there is no way around this one path. So please make it configurable. I could live with a hardcoded path, too, if it was one of the aforementioned plus an xorg subfolder.

hosxy:
I think so.     
`.xsession-errors` in `$HOME ` is not a good idea, I think it will be more suitable where it is in `$XDG_CACHE_HOME` or `$XDG_RUNTIME_DIR`

nefigtut:
+1 to this, it would be great to have a configurable location for .xsession-errors.
i ended up binary hacking /usr/sbin/lightdm to change the hardcoded name to the ".cache/xs-errors".

funny enough, this issue dates back to 2012 (if not earlier):
https://bugs.launchpad.net/ubuntu/+source/lightdm/+bug/1001035

Alexander-Shukaev:
This is a simple, but useful change.  Uncontrollable growth of `.xsession-errors` in `$HOME` and, in general, its presence there drive people nuts all over the web.  Please, fix this in the upcoming release.

haarp:
Quick'n'dirty hack:

```
--- a/src/session.c     2018-08-30 02:28:55.000000000 +0200
+++ b/src/session.c     2020-01-08 14:57:16.495681494 +0100
@@ -981,7 +981,7 @@
 {
     SessionPrivate *priv = session_get_instance_private (session);

-    priv->log_filename = g_strdup (".xsession-errors");
+    priv->log_filename = g_strdup (".cache/xsession-errors");
     priv->log_mode = LOG_MODE_BACKUP_AND_TRUNCATE;
     priv->to_child_input = -1;
     priv->from_child_output = -1;
```

doesn't follow `$XDG_CACHE_HOME` but works for me.

azureblue:
[...]

PR #287, opened 2023-01-11
Make location of .xsession-errors configurable

nefigtut:
LightDM always uses $HOME/.xsession-errors as the log file because it is harcoded. There are several possible reasons for someone to define another location, for example to write it into the RAM or to just have a tidier $HOME-folder.

This is a long-awaited feature, see: https://github.com/canonical/lightdm/issues/95 and https://bugs.launchpad.net/ubuntu/+source/lightdm/+bug/1001035

Unfortunately, I have no idea how to fix tests to conform this change. Please, feel free to update as needed.

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -981,7 +981,14 @@ session_init (Session *session)
 {
     SessionPrivate *priv = session_get_instance_private (session);
 
-    priv->log_filename = g_strdup (".xsession-errors");
+    if (g_getenv ("XDG_STATE_HOME"))
+        priv->log_filename = g_build_filename (g_getenv ("XDG_STATE_HOME"), ".xsession-errors", NULL);
+    else
+        if (g_getenv ("XDG_CACHE_HOME"))
+            priv->log_filename = g_build_filename (g_getenv ("XDG_STATE_HOME"), ".xsession-errors", NULL);
+        else
+            priv->log_filename = g_build_filename (".cache", ".xsession-errors", NULL);
+
     priv->log_mode = LOG_MODE_BACKUP_AND_TRUNCATE;
     priv->to_child_input = -1;
     priv->from_child_output = -1;


Soapux (commented):
I think it should be XDG_STATE_HOME or ~/.local/state if not found, which was designed for such log files.  Personally, I think it should probably go in it's on x11 sub directory too since more than one file is created, but it's not mandatory. So ~/.local/state/x11/xsession-errors in that case. I would also remove the dot from the file. Lastly we should probably check if the directories exist and if not, create them with mode 700 to follow the spec.

nefigtut:
no one from lightdm maintainers/devs has ever reacted to this pr or to #95.
while your suggestion is reasonable, i don't see any sense investing efforts into updating this pr knowing it will be ignored anyway. i give up and version-lock hacked lightdm on my systems.

src/session.c
```diff
@@ -981,7 +981,14 @@ session_init (Session *session)
 {
     SessionPrivate *priv = session_get_instance_private (session);
 
-    priv->log_filename = g_strdup (".xsession-errors");
+    if (g_getenv ("XDG_STATE_HOME"))
+        priv->log_filename = g_build_filename (g_getenv ("XDG_STATE_HOME"), ".xsession-errors", NULL);
+    else
+        if (g_getenv ("XDG_CACHE_HOME"))
+            priv->log_filename = g_build_filename (g_getenv ("XDG_STATE_HOME"), ".xsession-errors", NULL);
```
rozhuk-im, line 988: ```XDG_CACHE_HOME``` should be here.

rozhuk-im (changes requested):
copy-paste fix required

rozhuk-im:
I am not sure that this work: ```XDG_*``` - who set these vars?

madpilot78:
> I am not sure that this work: `XDG_*` - who set these vars?

@rozhuk-im It really depends. As far as I understand in most Linux distributions it is systemd setting them nowadays. But these can also be populated by PAM modules or session startup scripts.

In FreeBSD it looks like the magic resides in `/usr/local/etc/xdg`

So my guess is that depending on how they are setup these variable could not be correctly populated when the code actually executes.

nefigtut:
#335 is a nicer version of this indeed.

ahmubashshir (commented):
IMO, if the log files are in `$XDG_CACHE_HOME` or `$XDG_STATE_HOME`, `.` doesn't need to be in the filename.

PR #288, opened 2023-01-28
x-server: Avoid reusing the local X server if the hostname has changed

tintou:
If the hostname has changed while using a local seat, we will fail to connect and return to the greeter. Avoid this behavior by recreating the server.

### DIFF
--- a/src/x-server.c
+++ b/src/x-server.c
@@ -27,6 +27,9 @@ typedef struct
     /* Authority */
     XAuthority *authority;
 
+    /* Cached hostname for the authority */
+    gchar local_hostname[1024];
+
     /* Connection to this X server */
     xcb_connection_t *connection;
 } XServerPrivate;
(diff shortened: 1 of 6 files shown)


src/x-server.c
```diff
@@ -108,7 +126,17 @@ x_server_get_session_type (DisplayServer *server)
...
+    gchar actual_local_hostname[1024];
+
+    g_return_val_if_fail (server != NULL, FALSE);
+
+    if (priv->local_hostname[0] == '\0')
+        return TRUE;
+
+    gethostname (actual_local_hostname, 1024);
+    return g_strcmp0 (actual_local_hostname, priv->local_hostname) == 0;
+
```
robert-ancell, line 139: Extra line added here.

src/x-server.c
```diff
@@ -108,7 +126,17 @@ x_server_get_session_type (DisplayServer *server)
...
-    return TRUE;
+    XServerPrivate *priv = x_server_get_instance_private ((XServer*) server);
+    gchar actual_local_hostname[1024];
+
+    g_return_val_if_fail (server != NULL, FALSE);
+
+    if (priv->local_hostname[0] == '\0')
+        return TRUE;
+
+    gethostname (actual_local_hostname, 1024);
```
robert-ancell, line 137: Add a comment above here about why the hostname affects the ability to share this server.

tintou: done :)

robert-ancell (approved):

PR #300, opened 2023-04-19
Multiseat bug fixes, and some cleanups

rhansen:
This PR contains a few fixes that affect multiseat, especially with Wayland.  There are also several cleanups.

The important bits:
  * Initialize `WaylandSessionPrivate.vt` to -1 so that LightDM doesn't erroneously attempt to switch to VT 0 or set `XDG_VTNR` on non-`seat0` seats.
  * Use systemd-logind to discover the active session on non-`seat0` seats rather than always returning `NULL`.  This prevents LightDM from immediately restarting and switching to the greeter concurrently with a new Wayland session.
  * Disable user switching if logind says it isn't supported.  Note that logind v245 and earlier erroneously reports `CanMultiSession=no` when it is actually supported so I believe this will break user switching for such users.  See systemd/systemd#15337.

I believe the latter two together fix [LP#1371250](https://bugs.launchpad.net/lightdm/+bug/1371250) (at least I can't reproduce it).

Please do not squash this when merging—the commits are intended to be separate.  I will rebase these as needed to incorporate feedback or update to the latest `main`.


rhansen:
This PR now depends on the commits from PR #301 so I'm marking this as draft until that is merged.

src/seat.c
```diff
@@ -768,6 +767,8 @@ session_stopped_cb (Session *session, Seat *seat)
     if (session == priv->next_session)
         g_clear_object (&priv->next_session);
     if (session == priv->session_to_activate)
+        /* We were waiting for this session, but it didn't start :( */
```
robert-ancell, line 770: Can you either move the comment before the if as before or add braces to make it clear the g_clear_object is part of the branch.

rhansen: Done.

rhansen:
Thanks @robert-ancell!

PR #301, opened 2023-04-22
Fix Fedora and Ubuntu GitHub workflow tests, and other test/CI improvements

rhansen:
The most important commits are:

  * tests: Don't redirect `/tmp/dbus-*` to `$LIGHTDM_TEST_ROOT/tmp/dbus-*`
  * ci: Re-enable Ubuntu tests

The first of those two commits fixes the `test.yaml` GitHub workflow for both Fedora and Ubuntu.  I don't know why `dbus-daemon` is creating a Unix domain socket in the filesystem instead of creating an abstract socket.  Maybe some security policy is preventing abstract sockets from working in Docker on the runner?

The remaining commits are test/CI tweaks I made while troubleshooting that I thought were worth including.

Please do not squash this when merging—the commits are intended to be separate.  I will rebase these as needed to incorporate feedback or update to the latest `main`.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -10,23 +10,25 @@ jobs:
       fail-fast: false
       matrix:
         image:
-          - ubuntu:rolling
-          - fedora:latest
+          - name: ubuntu
+            tag: rolling
+          - name: fedora
+            tag: latest
 
     runs-on: ubuntu-latest
-    container: ${{ matrix.image }}
+    container: ${{ matrix.image.name }}:${{ matrix.image.tag }}
 
     steps:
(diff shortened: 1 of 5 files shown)


rhansen:
> I don't know why `dbus-daemon` is creating a Unix domain socket in the filesystem instead of creating an abstract socket.

Ah ha: https://gitlab.freedesktop.org/dbus/dbus/-/blob/35ade3c8f7aca16d1c6289828a2597859d1c503b/NEWS#L129-L147

robert-ancell:
Thanks @rhansen! I've never managed to find the time to fix this bitrot, much appreciated!

PR #323, opened 2023-09-21
Add support for changing the home directory in PAM modules

### LINKED ISSUE #322: LightDM should support PAM modules which change the home directory
PAM modules such as pam_mklocaluser may change or even create the home directory. Currently, LightDM assumes that the home directory will not change when opening the PAM session, the user's home directory is obtained via getpwent() after authentication but before opening the session.  In case of pam_mklocaluser this assumption is invalid, it creates a local user account and home directory from the user information that was obtained via LDAP and updates the HOME environment variable in PAM to the newly created directory which may differ from the value previously obtained from LDAP. The LightDM session child should obtain the location of the home directory from the PAM environment after opening the PAM session to support this.

A second, related problem is that if the Xauthority file is not stored in a system directory, the daemon hardcodes its path to the user's home directory and passes it as an absolute path to the session child. The final path should rather be constructed in the session child after opening the PAM session using the potentially changed value of the HOME environment variable.

gber:
PAM modules such as pam_mklocaluser may change or even create the home directory. Currently, LightDM assumes that the home directory will not change when opening the PAM session, the user's home directory is obtained via getpwent() after authentication but before opening the session.  Fix this by trying to update the user's home directory from the HOME environment variable from PAM after opening the session.
Furthermore, if the Xauthority file is not stored in a system directory, the daemon hardcodes its path to the user's home directory and passes it as an absolute path to the session child.  Fix this by passing it as a relative path so that the actual path can be constructed after the PAM session has been opened and the home directory has potentially been updated.

This fixes #322.

### DIFF
--- a/src/session-child.c
+++ b/src/session-child.c
@@ -543,6 +543,8 @@ session_child_run (int argc, char **argv)
     if (!home_directory) {
         home_directory = user_get_home_directory (user);
     }
+    if (version >= 4)
+        write_string (home_directory);
 
     /* Open a connection to the system bus for ConsoleKit - we must keep it open or CK will close the session */
     g_autoptr(GError) error = NULL;
--- a/src/session-child.c
+++ b/src/session-child.c
@@ -538,6 +538,12 @@ session_child_run (int argc, char **argv)
         return EXIT_FAILURE;
     }
 
+    /* try to get HOME from PAM since it might have been changed */
+    const gchar *home_directory = pam_getenv (pam_handle, "HOME");
+    if (!home_directory) {
+        home_directory = user_get_home_directory (user);
+    }
+
     /* Open a connection to the system bus for ConsoleKit - we must keep it open or CK will close the session */
     g_autoptr(GError) error = NULL;
     g_autoptr(GDBusConnection) bus = g_bus_get_sync (G_BUS_TYPE_SYSTEM, NULL, &error);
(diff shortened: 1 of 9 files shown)


src/session-child.c
```diff
@@ -651,8 +661,10 @@ session_child_run (int argc, char **argv)
         /* NOTE: This must be done after the permissions are changed because NFS filesystems can
          * be setup so the local root user accesses the NFS files as 'nobody'.  If the home directories
          * are not system readable then the chdir can fail */
-        if (chdir (home_directory) != 0)
+        if (chdir (home_directory) != 0) {
+            g_printerr ("chdir: %s\n", strerror (errno));
```
robert-ancell, line 665: Is this required? Looks like left-over debugging.

robert-ancell (changes requested):
This looks generally good. I think we'd want to return the home directory in the "Report authentication result" section in session-child.c so other code can access it (i.e. in the same way the username is returned).

It really should have a test in tests/, could you have a go at one?

gber:
> This looks generally good. I think we'd want to return the home directory in the "Report authentication result" section in session-child.c so other code can access it (i.e. in the same way the username is returned).

OK, I'm now passing it back and storing in the Session object. I've also versioned the change, please check whether it is correct like this.

I've also modified the script hooks running in the context of the user to use the home directory passed back to the daemon. 
 
> It really should have a test in tests/, could you have a go at one?

I have added a test case.

sunweaver:
@robert-ancell I checked the status of this PR today and noticed that a second review iteration is necessary. I'd really love to see this patch land in Debian soon, but for this we need upstream approval. Can you take a second look? Thanks!

robert-ancell (approved):
Thanks!

PR #326, opened 2023-10-20
fix(login): Fixed User desktop login failure after the.Xauthority file permission or authname is modified

winnerym100:
Fixed User desktop login failure after the.Xauthority file permission or authname is modified

Description: 修复用户的.Xauthority文件权限或authname被修改后无法登录用户桌面问题

Log: null

### DIFF
--- a/src/x-authority.c
+++ b/src/x-authority.c
@@ -294,7 +294,8 @@ x_authority_write (XAuthority *auth, XAuthWriteMode mode, const gchar *filename,
         if (!matched &&
             priv->family == a_priv->family &&
             address_matches &&
-            strcmp (priv->number, a_priv->number) == 0)
+            strcmp (priv->number, a_priv->number) == 0 &&
+            strcmp (priv->authorization_name, a_priv->authorization_name) == 0)
         {
             matched = TRUE;
             if (mode == XAUTH_WRITE_MODE_REMOVE)


src/x-authority.c
```diff
@@ -294,7 +294,8 @@ x_authority_write (XAuthority *auth, XAuthWriteMode mode, const gchar *filename,
         if (!matched &&
             priv->family == a_priv->family &&
             address_matches &&
-            strcmp (priv->number, a_priv->number) == 0)
+            strcmp (priv->number, a_priv->number) == 0 &&
+            strcmp (priv->authorization_name, a_priv->authorization_name) == 0)
```
robert-ancell, line 298: This change looks good, but it unrelated to the above change - can you do these as two separate PRs?

src/session-child.c
```diff
@@ -603,8 +603,13 @@ session_child_run (int argc, char **argv)
     if (x_authority)
     {
         gboolean drop_privileges = geteuid () == 0;
-        if (drop_privileges)
+        if (drop_privileges) {
+            if (x_authority_filename && chown (x_authority_filename, user_get_uid (user), user_get_gid (user)) < 0)
+                g_warning ("Failed to set ownership of user authority file: %s", strerror (errno));
             privileges_drop (user_get_uid (user), user_get_gid (user));
+            if (x_authority_filename && chmod (x_authority_filename, S_IRUSR | S_IWUSR) < 0)
```
robert-ancell, line 610: The chmod should be in x-authority.c I think - that's where it is set for newly created files.

src/x-authority.c
winnerym100: Okay, I will resubmit the PR separately. Thank you！

src/session-child.c
winnerym100: In the actual application process, an old file was accidentally modified with permissions, resulting in the inability to start a user session. This is done to remedy this abnormal situation.

Issue #269, opened 2022-09-23
Lightdm does not respect login.conf on FreeBSD


DanielO:
Hi,
I have ~/.login_conf to set some variables (eg language) which works via SSH but not lightdm - this is because lightdm does not call setusercontext.

I have a patch which calls it, although it is a bit of a kludge because setusercontext has no option to write into a PAM context.

https://gist.github.com/DanielO/178a3131cc1ae646e703d59e5d9c712d

Note that this patch is on top of the FreeBSD port patch, hence the removal of updwtmpx calls.

Is this something that would be upstreamable, or should I submit it as a FreeBSD port patch?

JackieMium:
This is a issue bothering me too. Hope the fix is merged. Thanks!

jpeisach:
Should have been fixed by that PR :)

PR #334, opened 2023-12-27
Use setusercontext(3) where available

madpilot78:
Hi,

I recreated the patch suggested by @DanielO in issue #269 as a pull request.

This patch adds checks in configure for setusercontext(3) and then uses it, if available, to switch user after forking int he session.

Using this method leverages existing FreeBSD code to se the user session that takes into account the login.conf values and also allows automated substitution of ~ (tilde) and $ (dollar) characters with their values.

This solves issues on FreeBSD and the patch is being discussed for inclusion in the FreeBSD port here: https://bugs.freebsd.org/266532

Manual page for setusercontext: [setusercontext(3)](https://man.freebsd.org/cgi/man.cgi?query=setusercontext&apropos=0&sektion=0&manpath=FreeBSD+14.0-RELEASE+and+Ports&arch=default&format=html)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -48,7 +48,7 @@ AC_CHECK_HEADERS(security/pam_appl.h, [], AC_MSG_ERROR(PAM not found))
 
 AC_CHECK_HEADERS(gcrypt.h, [], AC_MSG_ERROR(libgcrypt not found))
 
-AC_CHECK_FUNCS(setresgid setresuid clearenv __getgroups_chk)
+AC_CHECK_FUNCS(setresgid setresuid setusercontext clearenv __getgroups_chk)
 
 PKG_CHECK_MODULES(LIGHTDM, [
     glib-2.0 >= 2.44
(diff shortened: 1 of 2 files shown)


madpilot78:
Regarding the agreement, I only reworked an existing patch already sent by the user as a Pull Request, I cannot assert any copyright myself on this patch.

So I cannot sign any agreement stating such assertion regarding this.

I'm not sure what the policy should be in this case.

DanielO:
It is also largely copied from xdm in any case, so perhaps that complicates it more...

madpilot78:
To any project developers, I'd really like to help including this in the official port. But I'm not able to complete the CLA form, since it asks me to fill a field `Canonical Project Manager or contact`, where I don't know what to write into.

If you give me this information I'll be at least able to sign the CLA.

robert-ancell:
@madpilot78 you can set me as the contact in the form, thanks! This should be good to land after that's complete.

madpilot78:
@robert-ancell

Thanks! I just compiled and submitted the agreement online. I have found no easy way to rerun checks on this PR (and #335) but everything should be fine now.

PR #335, opened 2023-12-27
Make location of .xession-errors depend on user configuration

madpilot78:
I have extended on the patch in #287 by adding a configuration file variable for lightdm to control how the location of `.xsession-errors` is chosen.

I added a `smart-xsession-errors`, which is off by default, that allows to switch from the old behavior (putting it in `~/.xsession-errors`) to the algorith suggested in #287

This patch is being considered for inclusion in the freebsd ports tree for lightdm [1]. I decided to add a configuration variable and set it to false (implying old behavior), because I don't feel comfortable forcing a change of the location of this file on users. This due to the fact that the location of this file is historical behavior and expected by the users.

[1] https://bugs.freebsd.org/266532 and https://bugs.freebsd.org/275885


madpilot78:
As in #334 I cannot claim full copyright on this, since I took part of the code from #287.

Also, I am unable to complete the CLA form, it asks for a `Canonical Project Manager or contact`. What should I put in this field?

Thanks in advance.

robert-ancell:
@madpilot78 you can put me in the contact field, thanks!

marcos-aparicio:
Is there any update on merging this PR?

spikethehobbitmage:
Rather than trying to auto-detect, it would be better to make it a string option defaulting to the legacy value. That way users can set it to whatever they want.

Issue #337, opened 2023-12-28
LightDM failsafe handler service


wallentx:
I recently made a contribution to a project that would provide a failsafe handler to LightDM, where if Xorg failed to start, it would put the user in TTY2 to attempt to remedy the issue (as opposed to just leaving the user faced with a blacked out screen while the LightDM service tries, and fails to restart repeatedly). https://github.com/LukeShortCloud/winesapOS/commit/5d9b78c312ffe6073e7e3e9ad737f7252defacc0

I wondered if something like this would be beneficial upstream. I'm happy to put together a PR if the idea seems within the scope of LightDM's realm of responsibility.

PR #340, opened 2024-01-04
LightDM failsafe handler and restart directives

wallentx:
This PR is in reference to https://github.com/canonical/lightdm/issues/337

In a few different scenarios, I've experienced situations where a child process spawned by LightDM (such as x11) fails to start. As per the instructions in [lightdm.service](https://github.com/canonical/lightdm/blob/main/debian/lightdm.service#L12), upon failure, LightDM will `Restart=always`.
The problem that can arise from this is when one of those services is failing for a very specific reason that requires the user to make changes. In a scenario where such a service will fail 100% of the time, this will cause LightDM to `Restart=always` 100% of the time, which will eventually trigger systemd's rate-limiting mechanism, resulting in logs that appear like this:
```
systemd[1]: lightdm.service: Service hold-off time over, scheduling restart.
systemd[1]: Stopped LightDM.                        
systemd[1]: lightdm.service: Start request repeated too quickly.
systemd[1]: Failed to start LightDM.                
systemd[1]: lightdm.service: Unit entered failed state.       
systemd[1]: lightdm.service: Failed with result 'exit-code'. 
```

The problem is that, depending on the failure, the user might only see a black screen, with no indication that anything is wrong other than the screen being black and nothing loading. The user would only be able to discover the details in the aforementioned logs if they switched to a different TTY, logged in, and checked the logs. From there, they would be able to discover why the failure happened and fix the issue.

The changes that I am proposing are essentially a shortcut to what I just described.

The `lightdm.service` file now includes StartLimit IntervalSec and Burst values, a RestartSec rate, and an OnFailure action. The values I've proposed I _think_ make sense:
The service has a restart delay of 5 seconds in between resets, and if the service restarts more than 5 times within a 60s period, then the service will be in a failed state. In the event of it being in a failed state, it triggers the OnFailure action:[debian/lightdm-failure-handler.service](https://github.com/wallentx/lightdm/blob/801d1f0a45fdf141a6c92ea9316b5e2d5ea08bd7/debian/lightdm-failure-handler.service)

This service makes a copy of `/etc/issue` (the MOTD-like message you see above a TTY login) to `/etc/issue.restore`, and writes in a new message "hint" to `/etc/issue` that lets the user know what to do. It then changes the user to `TTY4` (I only chose 4 because I've noticed others to be reserved for other things, or unavailable.. someone should check my reasoning here though) so that they can see that message, and log in to troubleshoot.

Upon [successful login](https://github.com/wallentx/lightdm/blob/wallentx/lightdm-failsafe-handler/debian/lightdm-success-handler.service) of LightDM, if an `/etc/issue.restore` file exists, it will replace `/etc/issue` (which presumably has the "hint" we wrote earlier). If no .restore file is found, no action is taken.

I couldn't find anything else in the repo specifically that referenced the packaging or manifest of the .service files, but maybe that is handled elsewhere.

### DIFF
--- a/debian/lightdm-failure-handler.service
+++ b/debian/lightdm-failure-handler.service
@@ -1,0 +1,9 @@
+[Unit]
+Description=Handle LightDM failure
+
+[Service]
+Type=oneshot
+ExecStart=/bin/mv /etc/issue /etc/issue.restore; /bin/sh -c 'echo -e "LightDM has failed to start. You are now in a TTY for troubleshooting...\nYou can log in and check for details in logs." > /etc/issue; /usr/bin/chvt 4'
+
+[Install]
+WantedBy=multi-user.target
--- a/debian/lightdm-success-handler.service
+++ b/debian/lightdm-success-handler.service
@@ -5,7 +5,7 @@ After=lightdm.service
 
 [Service]
 Type=oneshot
-ExecStart=/bin/mv /etc/issue.restore /etc/issue
+ExecStart=/bin/sh -c 'test -f /etc/issue.restore && mv /etc/issue.restore /etc/issue || echo -e "Linux \\\\r (\\\\l)\n\n" > /etc/issue'
 
 [Install]
 WantedBy=graphical.target
--- a/debian/lightdm-success-handler.service
+++ b/debian/lightdm-success-handler.service
@@ -5,7 +5,7 @@ After=lightdm.service
 
 [Service]
 Type=oneshot
-ExecStart=/bin/sh -c 'test -f /etc/issue.restore && mv /etc/issue.restore /etc/issue || echo -e "Linux \\\\r (\\\\l)\n\n" > /etc/issue'
+ExecStart=/bin/sh -c 'test -f /etc/issue.restore && mv /etc/issue.restore /etc/issue || echo "File /etc/issue.restore not found. No restore action needed."'
 
 [Install]
 WantedBy=graphical.target
--- a/debian/lightdm-success-handler.service
+++ b/debian/lightdm-success-handler.service
@@ -1,0 +1,11 @@
+[Unit]
+Description=Reset /etc/issue after LightDM starts successfully
+Requires=lightdm.service
+After=lightdm.service
+
+[Service]
+Type=oneshot
+ExecStart=/bin/mv /etc/issue.restore /etc/issue
+
+[Install]
+WantedBy=graphical.target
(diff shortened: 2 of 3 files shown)


wallentx:
> Hey! wallentx has not signed the Canonical CLA which is required to get this contribution merged on this project.
> Please head over to [ubuntu.com/legal/contributors](https://ubuntu.com/legal/contributors) to read more about it.

I signed this, btw, but maybe my launchpad ID was supposed to be my launchpad account email?

robert-ancell:
Can you please resolve the above conflicts? Thanks.

robert-ancell:
Sorry, @wallentx needs a rebase again.

iphands:
This is great @wallentx I want this too!

Its rather maddening when you get caught in a restart loop and as you are at some tty, the restart occurs and in the middle of typing you get flipped back to tty1 as things fail.

Having some back off / retry limit would be great ootb config!

wallentx:
@robert-ancell I'm glad that the cla-check workflow is now passing, but to be honest, I'm not sure that it should be.
Prior to making this PR, I had went through all of the documented steps to bind my launchpad account to my github account, but prior executions of the cla-check action were failing:
https://github.com/canonical/lightdm/actions/runs/7636791934/job/20804451992?pr=340#step:2:50

The logs from that run seem to indicate to me that I either did some thing incorrect when setting up my launchpad account, or that there was some issue with launchpad and github being able to identify my details.

However after my rebase, the cla-check seems to have passed, and there is no evidence of my user being checked for a launchpad account:
https://github.com/canonical/lightdm/actions/runs/8576750571/job/23508258492?pr=340#step:2:48

I'm not sure if it is simply able to locate my launchpad account now, or if I was able to sneak in under the validation of these other users whose commits I pulled in from my rebase.

wallentx:
There's also a chance that I've made this PR ugly when rebasing, and I'm happy to just close this and redo it.

wallentx:
Hey @robert-ancell I've been held up on attempting to fix my Canonical CLA issue because I found a bug/exploit/bypass with the Canonical cla-check itself, and have a PR that shows it's existence, and also includes the fix https://github.com/canonical/has-signed-canonical-cla/pull/38

If I fix my CLA, then the evidence of the flaw disappears. Do you know anyone who might be able to at least ack the details of my other PR so that I can be freed up to move forward here?

robert-ancell:
@wallentx thanks for working on this! I've pinged the person who seems to be maintaining the CLA checker and hopefully they'll land that fix soon :)

PR #343, opened 2024-02-20
Implement logic to get FreeBSD ttyv device names.

madpilot78:
In FreeBSD the virtual terminals used for Xorg are named ttyv.

Their device numbers are also base32 encoded.

Code mostly taken from https://github.com/swaywm/wlroots/commit/fc6c0ca12e941d5d7d567834bff3ab7df9447001


madpilot78:
As an extra note, this patch is an improvement on a patch already present (and needed) the FreeBSD Ports Tree.

I'm planning to add this to the FreeBSD ports tree shortly.

robert-ancell (approved):
Thanks!

Issue #349, opened 2024-03-27
Let users force display number to match seat number


iphands:
I would like to have `seat0` always have `:0` and `seat1` have `:1`.
Right now it seems if I try and force this using any/all of:
- `xserver-command=/usr/bin/X :0`
- `xserver-display-number=0`

I end up with a situation like:
```
$ ps aux | fgrep X
root       21768  4.6  0.0 6652828 78060 ?       S<l  23:26   0:00 /usr/bin/X :1 :0 -layout seat1 -seat seat1 -auth /var/run/lightdm/root/:0 -nolisten tcp
root       21770 12.5  0.0 25396784 64132 tty7   S<sl+ 23:26   0:01 /usr/bin/X :0 :1 -layout seat0 -seat seat0 -auth /var/run/lightdm/root/:1 -nolisten tcp vt7 -novtswitch
```

I think this is due to how the code appends display_number no matter what.
https://github.com/canonical/lightdm/blob/main/src/x-server-local.c#L480

And that number is selected via:
https://github.com/canonical/lightdm/blob/main/src/x-server-local.c#L157

Maybe we could support a `forced_display_number` and fail if its unavailable?
Or even attempt to line up with the seat number?
Or attempt to line up with a `desired_display_number` and fall back to next available if that does not work?

iphands:
This was being looked for here: https://unix.stackexchange.com/questions/721870/force-seatn-to-display-n

iphands:
Oh I even tried this after reading the code:
```
[Seat:seat0]
xserver-command=/usr/bin/X :0
xserver-layout=seat0
xserver-display-number=0
minimum-display-number=0

[Seat:seat1]
xserver-command=/usr/bin/X :10
xserver-layout=seat1
xserver-display-number=10
minimum-display-number=10
```

But I guess it doesnt work (same `X :0 :1` and `X :1 :0` behavior) because `minimum-display-number` is not valid for an individual seat?

Its a global conf only option or something?

PR #350, opened 2024-03-28
Let multiseat X users choose a desired X display number

iphands:
This resolves issue #349 see details about what/why there.
If users supply the requested number we to use it _but_ move on with regular logic (next int) if we fail.


iphands:
I signed the CLA as I was opening the PR... perhaps it take a few mins to take effect?

PR #354, opened 2024-04-30
Switch back to the greeter when logind emits SecureAttentionKey 

n3rdopolis:
This depends on https://github.com/systemd/systemd/pull/29542 which is going to be included in the upcoming systemd v257 release as it was merged

### DIFF
--- a/src/lightdm.c
+++ b/src/lightdm.c
@@ -551,6 +551,13 @@ login1_service_seat_removed_cb (Login1Service *service, Login1Seat *login1_seat)
     remove_login1_seat (login1_seat);
 }
 
+static void
+login1_service_seat_attention_key_cb (Login1Service *service, Login1Seat *login1_seat)
+{
+   Seat *seat = display_manager_get_seat (display_manager, login1_seat_get_id (login1_seat));
+   seat_switch_to_greeter (seat);
+}
+
 int
 main (int argc, char **argv)
 {
@@ -891,6 +898,7 @@ main (int argc, char **argv)
         {
             g_signal_connect (login1_service_get_instance (), LOGIN1_SERVICE_SIGNAL_SEAT_ADDED, G_CALLBACK (login1_service_seat_added_cb), NULL);
             g_signal_connect (login1_service_get_instance (), LOGIN1_SERVICE_SIGNAL_SEAT_REMOVED, G_CALLBACK (login1_service_seat_removed_cb), NULL);
+            g_signal_connect (login1_service_get_instance (), LOGIN1_SERVICE_SIGNAL_SEAT_ATTENTION_KEY, G_CALLBACK (login1_service_seat_attention_key_cb), NULL);
 
             for (GList *link = login1_service_get_seats (login1_service_get_instance ()); link; link = link->next)
             {
--- a/src/login1.c
+++ b/src/login1.c
@@ -14,6 +14,7 @@
 #include <gio/gio.h>
 
 #include "login1.h"
+#include "seat.h"
 
 #define LOGIN1_SERVICE_NAME "org.freedesktop.login1"
 #define LOGIN1_OBJECT_NAME "/org/freedesktop/login1"
@@ -22,6 +23,7 @@
 enum {
     SEAT_ADDED,
     SEAT_REMOVED,
+    SEAT_ATTENTION_KEY,
     LAST_SERVICE_SIGNAL
 };
 static guint service_signals[LAST_SERVICE_SIGNAL] = { 0 };
@@ -247,6 +249,18 @@ signal_cb (GDBusConnection *connection,
             priv->seats = g_list_remove (priv->seats, seat);
             g_signal_emit (service, service_signals[SEAT_REMOVED], 0, seat);
         }
+    }
+    else if (strcmp (signal_name, "SecureAttentionKey") == 0)
+    {
+        const gchar *id, *path;
+        g_variant_get (parameters, "(&s&o)", &id, &path);
+
+        Login1Seat *seat = login1_service_get_seat (service, id);
+        if (seat)
+        {
+            g_signal_emit (service, service_signals[SEAT_ATTENTION_KEY], 0, seat);
+        }
+
     }
 }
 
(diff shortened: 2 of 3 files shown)


n3rdopolis:
The failure seems to be `test-power-qt5` now which seems to have also failed in another PR, so it doesn't appear related AFAIK, unless I am wrong

n3rdopolis:
https://github.com/systemd/systemd/pull/29542 is now merged for v257

n3rdopolis:
@robert-ancell the change that this depended on just landed in logind. not sure if you want to wait until v257 though for it to make it to a stable release or not?

n3rdopolis:
Let me know if I need to make any changes to this

n3rdopolis:
@robert-ancell Let me know if I need to make changes to get this merged, the feature that this is dependent on in systemd has been merged now

n3rdopolis:
@robert-ancell  I am still interested in getting this merged, is there anything I need to change please?

n3rdopolis:
Gentle ping, gdm just merged an equivalent patch earlier this week ;)

n3rdopolis:
@robert-ancell I would like a review if possible please

n3rdopolis:
@robert-ancell Gentle ping? Or are we waiting until systemd v257?

n3rdopolis:
@robert-ancell Systemd v257-rc1 has been tagged, do you think a review is possible please?

n3rdopolis:
@robert-ancell systemd v257 is officially released! Gentle ping?

n3rdopolis:
@robert-ancell Are you able to review this please? I'd like to get this merged now that the functionality is officially in a tagged systemd release. (v257)

n3rdopolis:
@robert-ancell  Gentle ping?

n3rdopolis:
Please let me know if I need to change anything?

n3rdopolis:
@robert-ancell Gentle ping? :(

n3rdopolis:
@robert-ancell should I force this to rerun again so it runs the new CLAv2 check?

jpeisach:
Hi @n3rdopolis :) this honestly looks good to me, I'm going to test it and then if it works I'll merge.

jpeisach:
Is there something I need to do in order to enable this behavior? I tested locally by merging into main and it's not doing anything (works on gdm3)

n3rdopolis:
I found out the problem. 
`can_multi_session` never gets set to true

So commit `fa2cf64a917d31605d40d34e98ce9e2e066064fa` in systemd, I submitted that since on VT-less seats, it would return CanMultiSession=no, even though, turns out there was no reason to

Later, in `8f8cc84ba4612e74cd1e26898c6816e6e60fc4e9` of systemd they dropped CanMultiSession totally from the properties list, kind of an odd choice IMO

But `add_seat` in `lightdm/src/login1.c` enumerates through logind properties of the seat, including CanMultiSession, since that property is not listed, even when I `busctl call org.freedesktop.login1 /org/freedesktop/login1/seat/seat0 org.freedesktop.DBus.Properties GetAll "s" ""`, CanMultiSession is never set to True. 

...so `seat_can_get_switch` does not really work, because that property is not being set to true.
Something like
```
diff --git a/src/login1.c b/src/login1.c
index 7fe84119..107ab6da 100644
--- a/src/login1.c
+++ b/src/login1.c
@@ -196,6 +196,7 @@ add_seat (Login1Service *service, const gchar *id, const gchar *path)
 
         const gchar *name;
         GVariant *value;
+        s_priv->can_multi_session = TRUE;
         while (g_variant_iter_loop (properties, "{&sv}", &name, &value))
         {
             if (strcmp (name, "CanGraphical") == 0 && g_variant_is_of_type (value, G_VARIANT_TYPE_BOOLEAN))
```
might be needed?

jpeisach:
Oof. I'm not a huge fan of doing that, especially since for a release this week I don't want any regressions. I also don't know how much this release would be back ported, and I don't want to potentially break systems (even if `can_multi_session` would be reset later to its proper value, if it's not listed and in a situation the answer should be `false`, I don't want it to accidentally trigger behavior).



I'm going to move this to the release afterwards so this change isn't mixed with bug fixes from up to 4 years ago. It will also solve the issue of systems that don't have systemd >= 257.

n3rdopolis:
OK, that is understandable since you're close to a release.


Just so you have background: `CanMultiSession` always reports true now since 2020 https://github.com/systemd/systemd/commit/fa2cf64a917d31605d40d34e98ce9e2e066064fa as there really is no case where it is actually supposed to be false.  ...and this change probably could have been made in systemd years before that, if it wasn't overlooked...

I first tested in 2017, and found that multisession on vt-less seats was actually working despite CanMultiSession=no.
I asked _why_ in 2020, like if I was overlooking some sort of behavior that that attribute was false but after I did, https://lists.freedesktop.org/archives/systemd-devel/2020-March/044201.html turns out it was an oversight, they added session switching (I think that mentioned work was done in 2013 if not 2014), but they forgot about that property.

A little while after I submitted that fix to make CanMultiSession=yes, they introduced another change to logind to make it not list the CanMultiSession attribute in the list of dbus properties, so since systemd v246, I think lightdm was assuming seats were CanMultiSession=no because of the change in logind...

So when the next release comes, I think assuming it to be true is fine probably.

n3rdopolis:
@JPeisach  Does this work now that the CanMultisession patch is merged?

I will admit, I don't know how to switch _back_ to the session though.
But I get the same behavior when I Switch User, which is the intended behavior of this patch.

jpeisach:
> @JPeisach Does this work now that the CanMultisession patch is merged?
> 
> I will admit, I don't know how to switch _back_ to the session though. But I get the same behavior when I Switch User, which is the intended behavior of this patch.

Yes, it does work, but my greeter (slick-greeter) offers the option to switch DE's.. Doing so doesn't do anything though, logging back in brings you back to the DE.

Actually, even when trying to switch user that still happened. So yes, the PR works as intended. I'm not merging yet because I am very likely to tag a release tonight or tomorrow for the CanMultiSession patch.

n3rdopolis:
makes sense! Thanks!

jpeisach:
I'll merge this in a day or two once the dust from 1.33.1 settles

Side note: It might be a fun idea to allow greeters to respond in their own way and have a custom screen that shows when `SecureAttentionKey` is emitted. It could help in cases when the desktop is entirely unresponsive.

I just need to check that supporting the feature would still keep LightDM.. light.

jpeisach:
Could you just rebase the branch to fix and retrigger the CI? Thanks

n3rdopolis:
OK, I hope I did this correctly. 
Really wish github rebased instead of making a merge commit... I hope this is correct

jpeisach:
I don't care, it's fine :)

PR #373, opened 2024-08-06
Add client library version for Qt6

Gliese852:
I would like to port `lightdm-kde-greeter` to qt6, and this requires the corresponding lightdm library.

I did not delete the version for qt5, the versions do not interfere with each other. The source code has not changed much, just updated the outdated `QString::toAscii` method.

Added an m4 macro to search for the `moc` executable file to check the version. It seems that this will be more reliable if there are several of them in the system. For reference, I used macros from the `freeciv` project. ([qt6.m4](https://github.com/freeciv/freeciv/blob/main/m4/qt6.m4))

### DIFF
--- a/m4/qt-validate-moc.m4
+++ b/m4/qt-validate-moc.m4
@@ -1,0 +1,68 @@
+# Detect Qt5 and Qt6 moc executable
+# partially borrowed from freeciv qt m4 scripts
+
+dnl Qt5
+AC_DEFUN([QT5_IF_QT5_MOC],
+    AS_IF([$1 -v >/dev/null 2>/dev/null &&
+        (test "`$1 -v 2<&1 | grep -o 'Qt [[[0-9]]]\+'`" = "Qt 5" ||
+         test "`$1 -v 2<&1 | grep -o 'moc [[[0-9]]]\+'`" = "moc 5" ||
+         test "`$1 -v 2<&1 | grep -o 'moc-qt[[[0-9]]]\+'`" = "moc-qt5")],
+        [$2]))
+
+dnl Set MOCCMD to $1 if it is the Qt 5 "moc".
+AC_DEFUN([QT5_TRY_MOC],
+    [QT5_IF_QT5_MOC([$1], [MOCCMD="$1"])])
+
+dnl If a usable moc command is found set $1
+AC_DEFUN([QT5_VALIDATE_MOC], [
+    AC_MSG_CHECKING([the Qt 5 moc command])
+
+    dnl Try to find a Qt 5 'moc'
+    AS_IF([test "x$MOCCMD" = "x"],
+        [for mocpath in "moc" "qtchooser -run-tool=moc -qt=5" "moc-qt5"
+        do
+            if test "x$MOCCMD" = "x" ; then
+                QT5_TRY_MOC([$mocpath])
+            fi
+        done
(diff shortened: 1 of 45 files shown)


Gliese852:
Well, now CI is building a version for Qt6. Tests fail just like in main; tests added for Qt6 also fail.

robert-ancell:
@Gliese852 yes, you will need to sign the CLA and you can use my name. Thanks!

Gliese852:
Squashed the fixup commits to at the same time provoke the launch of CI, at least the cla-check is now successful.

topazus:
Any updates with this PR?

xzl01:
Any updates?

Conan-Kudo:
It'd be great to see this land and released sooner rather than later, since the LightDM KDE Greeter requires this.

topazus:
> It'd be great to see this land and released sooner rather than later, since the LightDM KDE Greeter requires this.

Can lightdm package in Fedora adopted this PR as a patch to build against qt6? Since that OpenSUSE, ALT Linux added the qt6 library [1, 2]

[1] https://build.opensuse.org/projects/X11:Utilities/packages/lightdm/files/lightdm-1.32.0-qt6-library.patch?expand=1
[2] https://packages.altlinux.org/en/sisyphus/srpms/lightdm/specfiles/

m4/qt-validate-moc.m4
```diff
@@ -0,0 +1,68 @@
+# Detect Qt5 and Qt6 moc executable
+# partially borrowed from freeciv qt m4 scripts
```
jpeisach, line 2: My only concern is the copyright here.

The files from [freeciv](https://github.com/freeciv/freeciv/blob/main/m4/qt6.m4) do not have a copying message at the top. The other m4 files in lightdm do.

freeciv is GPL v2, which should make this okay.

But whether to include the comments at the top here, I'm not sure. I'm not knowledgeable about how to handle these kind of things.

tests/src/test-qt-greeter.cpp
```diff
@@ -73,11 +73,11 @@ void TestGreeter::printHints ()
     if (!showRemoteLoginHint ())
         status_notify ("%s SHOW-REMOTE-LOGIN-HINT=FALSE", greeter_id);
     if (autologinUserHint () != "")
-        status_notify ("%s AUTOLOGIN-USER-HINT=%s", greeter_id, autologinUserHint ().toAscii ().constData ());
+        status_notify ("%s AUTOLOGIN-USER-HINT=%s", greeter_id, autologinUserHint ().toLatin1 ().constData ());
     if (autologinGuestHint ())
         status_notify ("%s AUTOLOGIN-GUEST-HINT", greeter_id);
     if (autologinSessionHint () != "")
-        status_notify ("%s AUTOLOGIN-SESSION-HINT=%s", greeter_id, autologinSessionHint ().toAscii ().constData ());
```
jpeisach, line 80: I think the `toAscii` to `toLatin` changes should be ok, I don't think we need to worry about qt4. I also tested the PR by building with just qt5 installed and the tests passed, so this is no concern

JPeisach (commented):
Looks good - I guess packaging to be done in another PR. Just concerned about the licensing

jpeisach:
Per discussion on Matrix, licensing will be fixed after merge

PR #378, opened 2024-10-01
debian: Drop ExecPreStart hack

tintou:
All the other display managers have been migrated since a while now.

### DIFF
--- a/debian/lightdm.service
+++ b/debian/lightdm.service
@@ -8,9 +8,6 @@ StartLimitBurst=5
 OnFailure=lightdm-failure-handler.service
 
 [Service]
-# temporary safety check until all DMs are converted to correct
-# display-manager.service symlink handling
-ExecStartPre=/bin/sh -c '[ "$(basename $(cat /etc/X11/default-display-manager 2>/dev/null))" = "lightdm" ]'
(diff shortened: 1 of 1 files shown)


lo1ol:
I faced some problem with this check too, in project that install custom greeter on system

systemd control singularity of the display manager's using alias "display-manager":
```bash
systemctl status display-manager

# disable previous one
systemctl disable display-manager
systemctl enable lightdm

# systemd controls that you can't enable several display-managers
# this command fails:
systemctl enable gdm
```

So, this check really looks like over check

jpeisach:
> I faced some problem with this check too, in project that install custom greeter on system
> 
> systemd control singularity of the display manager's using alias "display-manager":
> 
> ```shell
> systemctl status display-manager
> 
> # disable previous one
> systemctl disable display-manager
> systemctl enable lightdm
> 
> # systemd controls that you can't enable several display-managers
> # this command fails:
> systemctl enable gdm
> ```
> 
> So, this check really looks like over check

sorry, can you clarify - does this PR cause a bug or no?

jpeisach:
Closing as we are about to remove the folder

lo1ol:
> sorry, can you clarify - does this PR cause a bug or no?

No, it fix it

PR #384, opened 2024-12-05
Rootless x

eylles:
yes this is #127 once again, attempting to continue from where @twhitehead left, will attempt to figure out how to make use of the `-displayfd` method, after all how hard could it be to check for a file descriptor am i right? /s

### DIFF
--- a/common/configuration.c
+++ b/common/configuration.c
@@ -399,6 +399,7 @@ config_init (Configuration *config)
 
     g_hash_table_insert (config->priv->vnc_keys, "enabled", GINT_TO_POINTER (KEY_SUPPORTED));
     g_hash_table_insert (config->priv->vnc_keys, "command", GINT_TO_POINTER (KEY_SUPPORTED));
+    g_hash_table_insert (config->priv->vnc_keys, "user", GINT_TO_POINTER (KEY_SUPPORTED));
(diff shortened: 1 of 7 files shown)


twhitehead:
I am okay with signing the Canonical CLA if you need me to to move this forward. (I would really like to see it make it in as well, just have been too busy to work on it more myself).

jpeisach:
> I am okay with signing the Canonical CLA if you need me to to move this forward. (I would really like to see it make it in as well, just have been too busy to work on it more myself).

The project has been moved to the Ubuntu GitHub organization and the Canonical CLA check has been removed, so you don't need to stress about this

@eylles any progress with this or have you abandoned the PR?

eylles:
sorry, i have not made any progress, wish i could say i didn't abandon the PR but i had to due to the difficulty on finding information about the display file descriptor that robert wanted to be used all the way back in 2020 since the socket polling can be flaky, https://github.com/ubuntu/lightdm/pull/127#pullrequestreview-452830679

i've been thinking of looking into the code of other display managers like sddm and gdm (well now it would have to be on the code for a release that did still support x11) to look at how they manage to do rootless x11 and shamelessly port said code... but i have not been able to do so

jpeisach:
ok, thanks. This is definitely a topic to come back to

PR #385, opened 2024-12-11
fix: fix some memory leak in session_child_run

KT-lcz:
The memory requested using g_strdup_printf is not released. Use g_autofree to release the memory.

### DIFF
--- a/src/session-child.c
+++ b/src/session-child.c
@@ -398,10 +398,14 @@ session_child_run (int argc, char **argv)
         {
             /* Set POSIX variables */
             pam_putenv (pam_handle, "PATH=/usr/local/bin:/usr/bin:/bin");
-            pam_putenv (pam_handle, g_strdup_printf ("USER=%s", username));
-            pam_putenv (pam_handle, g_strdup_printf ("LOGNAME=%s", username));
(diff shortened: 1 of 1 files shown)


src/session-child.c
```diff
@@ -398,10 +398,14 @@ session_child_run (int argc, char **argv)
         {
             /* Set POSIX variables */
             pam_putenv (pam_handle, "PATH=/usr/local/bin:/usr/bin:/bin");
-            pam_putenv (pam_handle, g_strdup_printf ("USER=%s", username));
-            pam_putenv (pam_handle, g_strdup_printf ("LOGNAME=%s", username));
-            pam_putenv (pam_handle, g_strdup_printf ("HOME=%s", user_get_home_directory (user)));
-            pam_putenv (pam_handle, g_strdup_printf ("SHELL=%s", user_get_shell (user)));
+            g_autofree char* user_env = g_strdup_printf ("USER=%s", username);
```
robert-ancell, line 401: For consistency char -> gchar, but otherwise everything looks great! Looks like putenv requires the string to remain, but pam_putenv makes a copy which is why this leak was probably added.

robert-ancell (approved):

src/session-child.c
KT-lcz: thanks you,I will correct the ‘char‘ to ‘gchar‘

PR #390, opened 2025-01-14
Allow users to set location of .xsession-errors

spikethehobbitmage:
A more flexible approach to #335 that allows users to set the .xsession-errors path directly. Supports shell expansion in the forms $VAR and ${VAR}.

Someone more experienced with glib should look it over. I'm not certain I've got the pointer semantics right. The GRegex object should probably be static, too.

### DIFF
--- a/common/configuration.c
+++ b/common/configuration.c
@@ -346,6 +346,7 @@ config_init (Configuration *config)
     g_hash_table_insert (config->priv->lightdm_keys, "greeters-directory", GINT_TO_POINTER (KEY_SUPPORTED));
     g_hash_table_insert (config->priv->lightdm_keys, "backup-logs", GINT_TO_POINTER (KEY_SUPPORTED));
     g_hash_table_insert (config->priv->lightdm_keys, "dbus-service", GINT_TO_POINTER (KEY_SUPPORTED));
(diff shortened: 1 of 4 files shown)


pkzc:
This seems to have stalled. Is there any way to nudge it on?

spikethehobbitmage:
@pkzc 

> This seems to have stalled. Is there any way to nudge it on?

I've stalled out. :( If someone else wants it, go ahead. It's a pretty trivial change but it probably needs to be rewritten. I don't claim any rights to it.

pkzc:
> @pkzc
> 
> > This seems to have stalled. Is there any way to nudge it on?
> 
> I've stalled out. :( If someone else wants it, go ahead. It's a pretty trivial change but it probably needs to be rewritten. I don't claim any rights to it.

Ah, I slipped over your second paragraph in the original post. Unfortunately I am not the one experienced in glib, let's hope somebody with knowledge  picks it up.

jpeisach:
I'm personally hesitant on the idea of using a regex, but I can think about it

There is also #338 which makes me wonder if this file should be in the same directory as .Xauthority, and therefore just have a custom XDG folder path?

PR #391, opened 2025-01-20
Adapt to Qt6

yixinshark:


JPeisach:
Hi! I think we are still requiring use of CLA - could you maybe sign it just in case?

debian/control
```diff
@@ -23,7 +23,7 @@ Build-Depends: autoconf,
                libxdmcp-dev,
                libxklavier-dev,
                pkg-config,
-               qtbase5-dev,
```
JPeisach, line 26: I don't know if we want to immediately remove qt5; it should eventually go away but I think it's best just to add the qt6 support instead of replacing it

JPeisach (changes requested):

JPeisach:
Also, how does this PR differ from #373?

PR #405, opened 2025-04-23
fix(seat init): Fix the issue of repeated initialization of seat0 cau…

winnerym100:
…sed by signal_cb callback probability executing in asynchronous threads

Description: Fix the issue of repeated initialization of seat0 caused by signal_cb callback probability executing in asynchronous threads

Log: null


jpeisach:
hi, do you have a way to reproduce the bug in the first place?

PR #406, opened 2025-04-23
fix(seat stop): when is stoping and the new seat not yet add to seats…

winnerym100:
…,need call stop the new seat

Description: when is stoping and the new seat not yet add to seats,need call stop the new seat

Log: null


jpeisach:
Is there a way to reproduce this?

PR #425, opened 2025-12-30
Fix the issue where the lightdm service is blocked after failing to c…

winnerym100:
…onnect to the X server upon receiving signal 10 from xorg, and add a retry mechanism


JPeisach:
Is there a way to reproduce the issue causing this in the first place?

PR #429, opened 2026-03-12
feat: add X server connection retry mechanism

chenjun1982:
Added configurable retry mechanism for X server connection failures when receiving SIGUSR1 signal. The system now supports retrying connection attempts with configurable timeout and maximum retry count. This improves reliability when X server startup encounters temporary issues.

New configuration options in lightdm.conf:
- xserver-retry-on-connection-failure: Enable/disable retry mechanism
- xserver-retry-timeout: Timeout in seconds for retry attempts (default:5)
- xserver-retry-max-count: Maximum retry attempts (default: 3)

The implementation tracks retry attempts, sets up timeout handlers, and properly cleans up resources. Connection success stops the retry mechanism immediately, while failures continue retrying until timeout or max attempts are reached.

Log: Added automatic retry for X server connection failures

Influence:
1. Test X server startup with normal successful connection
2. Test X server connection failure scenario with retry mechanism enabled
3. Verify retry timeout behavior with different timeout values
4. Test maximum retry count enforcement
5. Verify configuration options work correctly in lightdm.conf
6. Test retry mechanism disabled configuration
7. Verify resource cleanup after connection success/failure
8. Test additional SIGUSR1 signals after successful connection

### DIFF
--- a/data/lightdm.conf
+++ b/data/lightdm.conf
@@ -56,6 +56,9 @@
 # xserver-share = True if the X server is shared for both greeter and session
 # xserver-hostname = Hostname of X server (only for type=xremote)
 # xserver-display-number = Display number of X server (only for type=xremote)
+# xserver-retry-on-connection-failure = True to enable retry mechanism when X server connection fails on first SIGUSR1 signal
+# xserver-retry-timeout = Timeout in seconds for retry mechanism (default: 5)
+# xserver-retry-max-count = Maximum number of retry attempts when X server connection fails (default: 3)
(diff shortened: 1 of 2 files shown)


chenjun1982:
When the Linux system restarts, LightDM starts the X service. LightDM then connects to the X service after receiving signal 10 from it. At this point, the X service may reset and resend signal 10, but LightDM does not reprocess this signal, resulting in a black screen.
企业微信截图_17725044236670

PR #315, opened 2023-07-27
Allow Wayland sessions to start on seat0 on kernels with no VTs

n3rdopolis:
This is an attempt to fix #314

PR #430, opened 2026-03-13
Fix CI

Gliese852:
As far as I understand, after #315 a significant number of tests broke. This was mostly due to the fact that the `CanTTY` property introduced in the PR was not added to the test dbus interface. Also, recently the actions stopped running completely because `actions/upload-artifact@v3` was deprecated.

Updated to `actions/upload-artifact@v4` and fixed the problem with `can_tty`.

### DIFF
--- a/src/lightdm.c
+++ b/src/lightdm.c
@@ -26,6 +26,7 @@
 #include "display-manager-service.h"
 #include "xdmcp-server.h"
 #include "vnc-server.h"
+#include "vt.h"
 #include "seat-xdmcp-session.h"
 #include "seat-xvnc.h"
 #include "x-server.h"
@@ -918,6 +919,10 @@ main (int argc, char **argv)
             {
                 set_seat_properties (seat, NULL);
                 seat_set_property (seat, "exit-on-failure", "true");
+
+                /* in the absence of login1 we find out this using our own heuristics */
+                seat_set_can_tty (seat, vt_can_multi_seat ());
+
                 if (!display_manager_add_seat (display_manager, seat))
                     return EXIT_FAILURE;
             }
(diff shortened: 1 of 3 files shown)


chenjun1982 (approved):

tests/src/test-runner.c
```diff
@@ -1580,13 +1583,15 @@ add_login1_seat (GDBusConnection *connection, const gchar *id, gboolean emit_sig
     seat->path = g_strdup_printf ("/org/freedesktop/login1/seat/%s", seat->id);
     seat->can_graphical = TRUE;
     seat->can_multi_session = TRUE;
+    seat->can_tty = TRUE;
```
jpeisach, line 1586: So the tests do pass when this is set to true, but other than that, is there a reason this should be set to true by default?

src/lightdm.c
```diff
@@ -918,6 +919,10 @@ main (int argc, char **argv)
             {
                 set_seat_properties (seat, NULL);
                 seat_set_property (seat, "exit-on-failure", "true");
+
+                /* in the absence of login1 we find out this using our own heuristics */
+                seat_set_can_tty (seat, vt_can_multi_seat ());
```
jpeisach, line 924: Could this lead to false results (where CanTTY is set to false?)

I don't think not being multiseat implies not being able to TTY a single seat.

tests/src/test-runner.c
jpeisach: Based on the comment:
/* TRUE if this seat has TTYs */

So for a test runner this should be fine

src/lightdm.c
jpeisach: Uhhuh.

If CanTTY is basically "seat has a TTY", then the function internally checking if tty0 is accessible works.

So this works, but I feel like the function name is.. maybe slightly misleading?

jpeisach: Considering the function is basically the same, I'm going to close this. Probably when doing a brush over documentation and function names I'll pick this up.. hopefully I remember, but if not, I don't want to keep holding up everyone's CI for this.

PR #438, opened 2026-06-11
user: fix unintentional fall-through

ryonakano:
Fixes an issue where the warning message introduced by `G_OBJECT_WARN_INVALID_PROPERTY_ID` is shown when reading `is-locked` property.


jpeisach:
Good catch. Shouldn't affect anything :)

CLA check has been disabled, tests will be fixed in #430 (and locally I've tested both together and no tests fail).

PR #439, opened 2026-07-08
Deduplicate matching Xauthority records

zhangpan2001:
When replacing an Xauthority cookie, keep one matching record and remove duplicate matches. Unrelated records are preserved.

This avoids leaving stale cookies for the current display while keeping manually added Xauthority entries intact.

### DIFF
--- a/src/x-authority.c
+++ b/src/x-authority.c
@@ -261,7 +261,7 @@ x_authority_write (XAuthority *auth, XAuthWriteMode mode, const gchar *filename,
             g_warning ("Error reading existing Xauthority: %s", read_error->message);
     }
     GList *records = NULL;
-    gboolean matched = FALSE;
+    gboolean matching_record_found = FALSE;
     while (input_offset != input_length)
     {
         g_autoptr(XAuthority) a = g_object_new (X_AUTHORITY_TYPE, NULL);
@@ -290,24 +290,28 @@ x_authority_write (XAuthority *auth, XAuthWriteMode mode, const gchar *filename,
             address_matches = i == priv->address_length;
         }
 
-        /* If this record matches, then update or delete it */
-        if (!matched &&
-            priv->family == a_priv->family &&
-            address_matches &&
-            strcmp (priv->number, a_priv->number) == 0)
+        gboolean record_matches = priv->family == a_priv->family &&
+                                  address_matches &&
+                                  strcmp (priv->number, a_priv->number) == 0 &&
+                                  strcmp (priv->authorization_name, a_priv->authorization_name) == 0;
+
+        /* Remove all matching records, or update and keep only the first one */
+        if (record_matches)
         {
-            matched = TRUE;
-            if (mode == XAUTH_WRITE_MODE_REMOVE)
+            gboolean duplicate_record = matching_record_found;
+            matching_record_found = TRUE;
+
+            if (mode == XAUTH_WRITE_MODE_REMOVE || duplicate_record)
                 continue;
-            else
-                x_authority_set_authorization_data (a, priv->authorization_data, priv->authorization_data_length);
+
+            x_authority_set_authorization_data (a, priv->authorization_data, priv->authorization_data_length);
         }
 
         records = g_list_append (records, g_steal_pointer (&a));
     }
 
-    /* If didn't exist, then add a new one */
-    if (!matched)
+    /* If setting or replacing a missing record, then add a new one */
+    if (mode != XAUTH_WRITE_MODE_REMOVE && !matching_record_found)
         records = g_list_append (records, g_object_ref (auth));
 
     /* Write records back */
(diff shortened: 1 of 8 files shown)


zhangpan2001:
Could a maintainer please approve the remaining workflow to run? Thanks!

zhangpan2001:
@JPeisach  Could you please approve the workflow run? The CI failure was caused by a deprecated actions/upload-artifact@v3 which has been fixed by rebasing onto the latest main. Thanks!

jpeisach:
Sorry, queued them up. I'll check on this tomorrow.

zhangpan2001:
> Sorry, queued them up. I'll check on this tomorrow.

Thanks! I noticed most test failures (286/322) seem pre-existing — they fail with expected "XSERVER-0 START VT=7 SEAT=seat0", likely because the container lacks VT support.

src/x-authority.c
```diff
@@ -290,17 +290,24 @@ x_authority_write (XAuthority *auth, XAuthWriteMode mode, const gchar *filename,
...
+        gboolean record_matches = priv->family == a_priv->family &&
+                                  address_matches &&
+                                  strcmp (priv->number, a_priv->number) == 0;
+
         /* If this record matches, then update or delete it */
-        if (!matched &&
-            priv->family == a_priv->family &&
-            address_matches &&
-            strcmp (priv->number, a_priv->number) == 0)
+        if (record_matches)
```
jpeisach, line 299: Notetaking: `matched` was always set to FALSE, so removing it is fine

src/x-authority.c
```diff
@@ -290,17 +290,24 @@ x_authority_write (XAuthority *auth, XAuthWriteMode mode, const gchar *filename,
...
         {
-            matched = TRUE;
             if (mode == XAUTH_WRITE_MODE_REMOVE)
+            {
+                matched = TRUE;
+                continue;
+            }
+
+            if (matched)
```
jpeisach: If mode is not XAUTH_WRITE_MODE_REMOVE, this would evaluate to `if(false)`, so this code would never run

src/x-authority.c
```diff
@@ -290,17 +290,24 @@ x_authority_write (XAuthority *auth, XAuthWriteMode mode, const gchar *filename,
...
-        if (!matched &&
-            priv->family == a_priv->family &&
-            address_matches &&
-            strcmp (priv->number, a_priv->number) == 0)
+        if (record_matches)
         {
-            matched = TRUE;
             if (mode == XAUTH_WRITE_MODE_REMOVE)
+            {
+                matched = TRUE;
```
jpeisach: It was already matched by the top of the condition (after `if(record_matches)`), so shouldn't it be defined up there?

JPeisach (commented):
#430 should fix the tests.

Is there a way to reproduce the issue that would cause "duplicate matching Xauthority records"?

I'll check the tests in another review.

src/x-authority.c
zhangpan2001: Right, the old block could only be entered while matched was false. I also renamed it to matching_record_found to make the state clearer.

src/x-authority.c
zhangpan2001: It is false for the first match, then stays true for the rest of the loop. In replace mode, this skips the second and later matching records. I rewrote it using duplicate_record to make that clearer.

src/x-authority.c
zhangpan2001: Done. The code now saves the previous value in duplicate_record, then sets matching_record_found at the top of the block. This keeps the first match and skips later ones.

jpeisach:
The PR honestly looks fine to me but I'm going to wait for others to review.

zhangpan2001:
> The PR honestly looks fine to me but I'm going to wait for others to review.

Thanks for the review!

Franklin-Qi:
LGTM. Please resolve the error message in the test check. @zhangpan2001

zhangpan2001:
> LGTM. Please resolve the error message in the test check. @zhangpan2001

@Franklin-Qi The failure is caused by the missing `CanTTY` property in the test login1 mock, which is fixed by [#430](https://github.com/ubuntu/lightdm/pull/430). But it has not been merged yet.

jpeisach:
Update your branch for the CI fix :)

zhangpan2001:
@JPeisach Done, thanks.

jpeisach:
my brain is still spinning, I feel like the code is accomplishing the same thing as before with no changes, just some moved variables and conditions, but the test does fail when I test locally.

I am just concerned about regression risk

zhangpan2001:
> my brain is still spinning, I feel like the code is accomplishing the same thing as before with no changes, just some moved variables and conditions, but the test does fail when I test locally.
> 
> I am just concerned about regression risk

I checked this locally. The current code passes all five unit tests and `test-stale-xauthority`. With the old code, `matched` becomes true after the first match, so later matching records are written back unchanged. The new code drops those duplicates. That is the behavior covered by the failing test.

jpeisach:
I'm going to leave this for 1.34.0

PR #445, opened 2026-08-02
reduce ProcessPrivate 80 to 64 bytes using align and replace gboolean to bool

GermanAizek:
@JPeisach congratulations, I paid attention to the code again and accidentally noticed that you used a very non-optimized glib gboolean, you can align booleans and reduce the size of the structure to 1 cache line.

More info about gboolean: https://docs.gtk.org/glib/types.html#gboolean

```
There is no validation when assigning to a gboolean variable and so it could
contain any value represented by a gint.
```

### DIFF
--- a/src/process.c
+++ b/src/process.c
@@ -38,14 +38,17 @@ typedef struct
 
     /* File to log to */
     gchar *log_file;
-    gboolean log_stdout;
-    LogMode log_mode;
+    LogMode log_mode; // enum range [-1, 0, 1]
+    gboolean log_stdout : 1; // bool range [0, 1]
+
+    /* TRUE to clear the environment in this process */
+    gboolean clear_environment : 1; // bool range [0, 1]
+
+    /* TRUE if stopping this process (waiting for child process to stop) */
+    gboolean stopping : 1; // bool range [0, 1]
 
     /* Command to run */
     gchar *command;
-
-    /* TRUE to clear the environment in this process */
-    gboolean clear_environment;
 
     /* Environment variables to set */
     GHashTable *env;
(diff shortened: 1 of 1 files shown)


GermanAizek:
@JPeisach final mem layout sizes
```cpp
typedef struct
{
    /* Function to run inside subprocess before exec */
    ProcessRunFunc run_func; // 8 byte
    gpointer run_func_data; // 8 byte

    /* File to log to */
    gchar *log_file; // 8 byte
    LogMode log_mode; // 4 byte (enum)
    bool log_stdout; // 1 byte

    /* TRUE to clear the environment in this process */
    bool clear_environment; // 1 byte

    /* TRUE if stopping this process (waiting for child process to stop) */
    bool stopping; // 1 byte

    /* Command to run */
    gchar *command; // 8 byte

    /* Environment variables to set */
    GHashTable *env; // 8 byte

    /* Process ID */
    GPid pid; // 4 byte

    /* Exit status of process */
    int exit_status; // 4 byte

    /* Timeout waiting for process to quit */
    guint quit_timeout; // 4 byte

    /* Watch on process */
    guint watch; // 4 byte
} ProcessPrivate;
```

jpeisach:
Generally the code has used the Glib Types in The Great Before (before I stepped up to maintain a few weeks ago), so I'm not sure how I feel about changing it. Everything else uses the non-Glib types anyway. Maybe there's a reason to keep it that way? But maybe it's worth changing for alignment..

GermanAizek:
> Maybe there's a reason to keep it that way? 

Can save glib gboolean members  using bit fields trim up to 1 bit or 1 byte

GermanAizek:
@JPeisach 

how do you like this change while maintaining compatibility with gboolean?

https://github.com/ubuntu/lightdm/pull/445/changes/ecb3e414e45fb1eb9df7efe11a7d7fc09140b826

PR #446, opened 2026-08-02
lightdm: move out of loop strlen call

GermanAizek:
@JPeisach 

Fix common C/C++ compiler flaw is that it doesn't optimize strlen calls inside loops.

Reference:
- https://stackoverflow.com/questions/78827988/why-doesnt-the-compiler-optimize-strlen-calls-in-a-loop-despite-it-being-a-p

Since I started contributing to heavily optimized AV1 decoder ([ffmpeg dav1d](http://aomedia.org/blog%20posts/What-is-dav1d-and-Why-it-Matters/)), there are high performance requirements. I experimentally found out that putting calculated expressions from the body of the loop into separate constants gives a strong two-fold increase. What is strange is the C/C++ compiler that GCC and Clang on -O3 cannot automatically optimize.

It's also visually more readable with a separate const variable.

-  https://code.videolan.org/videolan/dav1d/-/merge_requests/1896
-  https://code.videolan.org/videolan/dav1d/-/merge_requests/1897
-  https://code.videolan.org/videolan/dav1d/-/merge_requests/1898
-  https://code.videolan.org/videolan/dav1d/-/merge_requests/1903

### DIFF
--- a/liblightdm-gobject/session.c
+++ b/liblightdm-gobject/session.c
@@ -122,6 +122,8 @@ load_sessions_dir (GList *sessions, const gchar *sessions_dir, const gchar *defa
     if (!directory)
         return sessions;
 
+    const size_t ext_len = strlen(".desktop");
+
     while (TRUE)
     {
         const gchar *filename = g_dir_read_name (directory);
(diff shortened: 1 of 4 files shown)


jpeisach:
I don't see any harm doing this - we don't want to compute strlen multiple times anyway

tintou:
Here we were doing strlen of constant expression, there shouldn't be any difference, the example you're linking is not the same as this, the only real occurrence is at src/session.c

jpeisach:
it's still a constant calculated expression. Also I imagine on builds with no compiler optimization at all, this would prevent the call from being made each time there is a .desktop file.

PR #449, opened 2026-08-02
xserver: fix incorrect format and x addr maybe null

GermanAizek:
x_server_get_address (server) may return NULL this dangerous since many functions may not check the input parameters for NULL

Affected funcs with maybe NULL parameter:
- l_debug
- xcb_connect_to_display_with_auth_info

### DIFF
--- a/src/x-server.c
+++ b/src/x-server.c
@@ -74,9 +74,9 @@ x_server_get_address (XServer *server)
     if (!priv->address)
     {
         if (priv->hostname)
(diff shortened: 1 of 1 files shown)


GermanAizek:
This cannot be fixed added check on NULL because g_strdup_printf(), g_snprintf() function belong to Glib.
l_debug fix He'll be here https://github.com/ubuntu/lightdm/pull/451

jpeisach:
No, like, if the server address is NULL then there is no server... so no connection should be attempted at that point

PR #450, opened 2026-08-02
xserver-local: fix incorrect format types

GermanAizek:
- display_number -> guint
- xdmcp_port -> guint
- x_server_get_display_number() return -> guint


jpeisach:
I'm surprised the compiler didn't throw a message for this. But again, it is hiding behind that type

PR #452, opened 2026-08-02
session-child: g_printerr/g_strdup_printf params may be null

GermanAizek:
Params:
- username
- x_authority_filename


GermanAizek:
this cannot be fixed added check on NULL because these functions belong to Glib

PR #453, opened 2026-08-03
seat: l_debug/g_printf params may be null

GermanAizek:
Params:
- name
- session_name


GermanAizek:
This cannot be fixed added check on NULL because g_printf() function belong to Glib.
l_debug fix He'll be here https://github.com/ubuntu/lightdm/pull/451

PR #454, opened 2026-08-06
Assume that the seat CanMultiSession by default

n3rdopolis:
logind having dropped listing the CanMultiSession property in 2024 now makes lightdm assume it is always false. This is incorrect as the reason why logind now hides this property, is because it is actually unconditionally true.

### DIFF
--- a/src/login1.c
+++ b/src/login1.c
@@ -196,6 +196,7 @@ add_seat (Login1Service *service, const gchar *id, const gchar *path)
 
         const gchar *name;
         GVariant *value;
+        s_priv->can_multi_session = TRUE;
         while (g_variant_iter_loop (properties, "{&sv}", &name, &value))
         {
             if (strcmp (name, "CanGraphical") == 0 && g_variant_is_of_type (value, G_VARIANT_TYPE_BOOLEAN))


jpeisach:
This will be basically the first thing merged for 1.34.0

jpeisach:
Actually can't the can_multi_session field just be dropped entirely across lightdm?

n3rdopolis:
Honestly? probably. I'd have to grep for everything though, ...do you want to leave in `allow-user-switching` from the config file though? I assume that's there for admins that only want one session for some reason

jpeisach:
uhhhhhhh I don't know, I'll get back to you on that.

jpeisach:
Yeah, keep the `allow-user-switching` option. Just remove the `can_multi_session` variable and check from the DBus properties

n3rdopolis:
I took it out of logind, but not out of everything, seems that seat-xvnc sets seat_set_supports_multi_session to false, which is not logind based, but that is still why I left the property and function in seat.c

jpeisach:
Will test when I get home

jpeisach:
LGTM, both code and in test.

jpeisach:
Oh, is there a new minimum version of systemd this would make lightdm require?

iczelia:
> Oh, is there a new minimum version of systemd this would make lightdm require?

No, I think that it's good as is.

n3rdopolis:
> Oh, is there a new minimum version of systemd this would make lightdm require?

Technically yes, but its very very veeeerry old. 

It seems to be systemd v209 from 2014
https://github.com/systemd/systemd/blob/65537f059c24bcaba6e70afb7767e67be62b5738/NEWS#L20088
that added multi-session on VT-less seats. So there likely is nothing shipping with systemd v208 out there...



What the problem is is that when they added that feature, they never fixed the CanMultisession attribute. It always returned no on VT-less seats, even after the change.

So for all these years, logind was telling login managers, display managers, and display servers that VT-less seats could not handle multisession, event though they could. 

No one really realized that property was wrong I guess, multi-seat is fairly rare. I fixed it in systemd in 2020, but I first suspected that it was wrong when I was testing in 2017. (I assumed it was like 90% complete or something, and not just a bug).

But yeah, then they made CanMultisession not get reported by ListProperties since it is always yes, which seems like that broke lightdm's enumeration of the properties, and it was not getting set to true at all. Even on VT-enabled seat0s

So I don't think you need to worry about bumping the system required version

PR #455, opened 2026-08-07
i18n: Small corrections in Occitan

Mejans:
A common mistake and some non breakable spaces


jpeisach:
Well, I don't think anyone ran `make update-po`. All of the translations appear to be for the LightDM GTK Greeter which was moved out of lightdm itself in https://github.com/ubuntu/lightdm/commit/c701addb2fac3016d022bf5bbebffcf8b086960a

So, as a result, all translations need to be cleaned up :/

Leaving this open, but it won't get merged right now.

Mejans:
Hello,
I'll update this PR when the other PR with updated data is merged ;)

PR #457, opened 2026-08-10
Remove debian/ folder

### LINKED ISSUE #456: Keep debian/ folder?
In ye olden days (before I became a maintainer and when the project was supposed by Canonical), the `debian` folder was where all Ubuntu packaging was done. This included patches for Ubuntu specifically.

However, fast forward to the previous release, 1.32.0. That release was not uploaded to Ubuntu. (That is why it is marked as UNRELEASED in `debian/changelog`). It was synced from Debian, and then Ubuntu placed its own patches on top of it for itself over time.

I do not have upload rights to Ubuntu, though I am sure I can ask for a sponsor if I need to get LightDM 1.33.0 in the archive. However, I wonder if it may be less redundant to have Debian package it, and then it gets synced to Ubuntu.

In that case, I also wonder if it's best to just remove the Debian folder entirely. This would make lightdm more distro-neutral, and prevents lightdm itself from needing to keep up with Debian packaging standards.

So before I start merging in changes for 1.34.0, I am thinking about how to handle this. In most releases, the main release commit would be made (the "Releasing $version"), and then another commit for the Ubuntu release would be made that just updates the debian/ folder.

Choices are:

1) Create a debian/changelog entry for 1.33.0, mark it as UNRELEASED
2) Import patches and packaging updates from Ubuntu into the debian/ folder, create a debian/changelog entry for 1.33.0 and mark it as UNRELEASED
3) Import patches and packaging updates from Ubuntu into the debian/ folder, create a debian/changelog entry for 1.33.0, mark it as RELEASED and get it sponsored into stonking
4) Remove the debian folder entirely

Option 1 makes it easy for anyone to `dpkg-buildpackage` after a clone (once they create an .orig tarball), but is essentially updating the changelog by copying the `NEWS` file.  Debian would update, Ubuntu would sync from Debian and place patches on top.

Option 2 makes anyone who clones the repository and runs `dpkg-buildpackage` to build basically a _predicted_ "Ubuntu" version of the package. Debian would then package for themselves, it gets synced to Ubuntu and Ubuntu adds patches as necessary.

Option 3 is the same as option 2, but the patched "Ubuntu" version is uploaded to Ubuntu, and Debian would package for themselves with their own patches, and hopefully it doesn't get synced to Ubuntu by accident and require replacing all the patches.

Option 4 means that, realistically, Debian would package, Ubuntu would sync and patch for themselves.

The first two options make the purpose of the debian folder.. like.. a suggestion?

The first three options make building a deb from a git clone easy, but anyone who does so will basically get an Ubuntu version, potentially breaking their distro's own patches.

The last option makes lightdm a "distro neutral" piece of software.

I need to check on the Ubuntu patches to see what they add. But before I create commits for 1.34.0, I'd like a "checkpoint" for Ubuntu to have its package. That may mean having a "1.33.0-ubuntu" branch.

jpeisach:
This allows for the project to be distro neutral, and prevents redundancy.

Closes #456

--------------------

Just need to check to make sure any important files aren't being removed, then probably going to leave it open for another day in case of any objections.

Reference package sources:
Fedora Rawhide - https://src.fedoraproject.org/rpms/lightdm/tree/rawhide
Debian - https://salsa.debian.org/xfce-extras-team/lightdm/-/tree/debian/master/debian
Arch - https://gitlab.archlinux.org/archlinux/packaging/packages/lightdm


debian/50-disable-guest.conf
```diff
@@ -1,5 +0,0 @@
-# Disable guest sessions due to them not being confined in systemd
-# CVE-2017-8900
-# https://bugs.launchpad.net/bugs/1663157
```
jpeisach, line 3: does this need to be an upstream default?

debian/patches/04_language_handling.patch
```diff
@@ -1,83 +0,0 @@
...
-Index: lightdm/liblightdm-gobject/language.c
-===================================================================
---- lightdm.orig/liblightdm-gobject/language.c
-+++ lightdm/liblightdm-gobject/language.c
-@@ -63,7 +63,7 @@ update_languages (void)
-     if (have_languages)
-         return;
- 
--    const gchar *command = "locale -a";
-+    const gchar *command = "/usr/share/language-tools/language-options";
```
jpeisach, line 14: I don't think this patch should be applied upstream. Honestly seems fine as is and, well, debian and ubuntu have managed without it for a couple of years

debian/patches/autologin-session-workaround.patch
```diff
@@ -1,22 +0,0 @@
-Description: Workaround faulty setting from Ubuntu user-setup package
-Author: Robert Ancell <robert.ancell@canonical.com>
-Bug: https://bugs.launchpad.net/bugs/1484083
```
jpeisach, line 3: maybe this should be sent upstream?

debian/lightdm.service
```diff
@@ -1,17 +0,0 @@
-[Unit]
```
jpeisach, line 1: should sample service files be shipped upstream?

Franklin-Qi:
The overall direction of this PR looks right to me: removing the in-tree debian/ directory keeps LightDM upstream distro-neutral.
  Before merge, I’d suggest adding two clarifications:
  1. explicitly state that the removed files are Debian/Ubuntu packaging assets only and do not affect upstream runtime code, tests, or core guest-session logic;
  2. if any patch under debian/patches still has upstream value, track it in a separate issue/PR instead of letting it disappear implicitly with this directory removal.

Also, please note that this overlaps with PRs still touching debian/ (for example #369), so those follow-up branches will need rebasing or cleanup.

debian/50-disable-guest.conf
Franklin-Qi: Not necessarily. If this behavior is primarily for Debian/Ubuntu packaging or downstream integration, I do not think it needs to be an upstream default. Upstream should stay distro-neutral and only carry defaults that make sense across distributions. If a distro wants this behavior by default, it is better handled in downstream packaging or config snippets rather than in the core upstream tree.

debian/patches/04_language_handling.patch
Franklin-Qi: Yeah, agreed. If Debian and Ubuntu have managed without it for a couple of years, I don't think it needs to live upstream. I'd keep this PR focused on removing debian/ and only revisit that patch separately if there's a current bug it still fixes

debian/50-disable-guest.conf
jpeisach: agreed :)

jpeisach:
#458 will track removed packaging defaults

PR #461, opened 2026-08-13
On FreeBSD there are always multiple VTs

madpilot78:
This patch is now required on FreeBSD. Before the `vt_can_multi_seat()` function was unused, and this was not noticed.

The patch was provided by Tĳl Coosemans <tijl@FreeBSD.org>


jpeisach:
That's.. it? You don't want to check any files for yourself?

madpilot78:
The FreeBSD kernel includes the vt driver by default and that driver is always able to provide multiple ttys.

I'm not even sure if a functional kernel can be built without the vt driver, but such a kernel would not be able to run Xorg or wayland or any king of graphic system anyway.

So, there is little to lookup. If the kernel is able to run Xorg or wayland multiple ttys are a possibility.

The user could configure just one vty, but the driver is anyway able to handle more.

This patch makes polkitd work correctly on FreeBSD, it failed if the function returned false.

jpeisach:
I'll trust you guys on this. I don't use FreeBSD so.. yeah

PR #464, opened 2026-08-18
liblightdm-gobject: define common-user GIR type

### LINKED ISSUE #463: liblightdm-gobject gir: "common-user" property missing name attribute in <type>
While working on [lightdm-rs](https://github.com/jpeisach/lightdm-rs), this message popped up:

```
[ERROR libgir::parser] Skipping property "common-user": Missing name attribute on <type> element
```

Indeed, the gir file does not assign common-user a type like others:

```
      <property name="background"
                transfer-ownership="none"
                getter="get_background"
                default-value="NULL">
        <type name="utf8" c:type="gchar*"/>
      </property>
      <property name="common-user"
                introspectable="0"
                readable="0"
                writable="1"
                construct-only="1"
                transfer-ownership="none">
        <type/>
```

In the code, this is a custom GObject, but I guess it's not being picked up because CommonUser (I think?) is defined in the ../common section of the code.

ousamabenyounes:
## Summary

- annotate the internal `common-user` property with its concrete introspection type
- reject generated GIR files that contain unnamed `<type>` elements when `xmllint` is available
- detect `xmllint` at configure time and distinguish inspection errors from malformed type metadata

## Test verification (RED → GREEN)

- RED check (same check with the production annotation removed): 1/1 new checks failed:

  ```text
  $ make check DEBUG=true
  LightDM-1.gir contains an unnamed type
  make[2]: *** [Makefile:1039: check-local] Error 1
  ```

- GREEN check (same check with the annotation restored): 1/1 new checks passed:

  ```text
  $ make check DEBUG=true
  Testsuite summary for lightdm 1.33.1
  # TOTAL: 352
  # PASS:  352
  # FAIL:  0
  # ERROR: 0
  ```

- pinned downstream parser reproduction: gtk-rs `gir` at `2ae5101` reports `Missing name attribute on <type> element` before the fix and no error after it
- full local suite, Ubuntu rolling: 352/352 passed
- full local suite, Fedora latest: 352/352 passed

Fixes #463


jpeisach:
Hi,

I personally am a little uneasy about vibe coded contributions - I hope you understand that.

I am not going to close the PR outright, since that would be rude, and likely violate [Ubuntu Code of Conduct](https://ubuntu.com/community/docs/ethos/code-of-conduct). In the meantime, I will get some other opinions about a contribution policy to lightdm, and acceptable uses of LLMs.

I also am not sure that the PR would actually fix the problem, and I would go into detail, but again I'm not sure how I feel about it due to potential use of training.

I will get back to you. Let me know if you need anything.

Thanks :)

jpeisach:
Hm.. I'm not sure if the xmllint is needed because the .gir files are generated automatically. I guess perhaps they could be a way of checking the output of the girs?

As for the type: It's now removed, and the `gir` tool isn't generating it... weird...

jpeisach:
Well, it is internal anyway.

I'm going to hold onto this until I figure out whether it should or should not include the property at all if it's unused

Issue #435, opened 2026-05-08
LightDM 1.32.0 crashes on XDMCP unmanaged-session timeout (double free)


agnon:
We are seeing a regression in LightDM on Debian Trixie.
   
## Environment
   
- Debian 13 (Trixie)
- lightdm 1.32.0-6+b2
- lightdm-gtk-greeter 2.0.9-1
   
## Setup
   
We have used this setup for a long time without problems on Bullseye:
   
- `xvnc.socket` listens on TCP 5900
- `xvnc@.service` starts:
  `Xvnc -inetd -query localhost -once -geometry 1920x1200 -SecurityTypes=TLSNone`
- LightDM has XDMCP enabled
   
A remote connection to TCP 5900 causes systemd to start Xvnc, which then makes a local XDMCP request to `localhost:177`.
   
## Actual result
   
On Trixie, LightDM crashes and systemd restarts it, which resets the active desktop session.
   
Journal shows:
   
```text
g_object_unref: assertion 'G_IS_OBJECT (object)' failed
double free or corruption (fasttop)
Source ID 30 was not found when attempting to remove it
lightdm.service: Main process exited, code=killed, status=6/ABRT
```

LightDM log shows the crash happens after an unmanaged XDMCP session times out:
```text
Got Query(authentication_names=[]) from 127.0.0.1:63690
Got Request(display_number=1 ...) from 127.0.0.1:63690
Send Accept(session_id=53140 ...) to 127.0.0.1:63690
Timing out unmanaged session 53140
```

## Expected result

LightDM should not crash when an unmanaged XDMCP session times out. It should clean up the session and continue running.

Regression

This same Xvnc/XDMCP arrangement worked on Debian Bullseye with LightDM 1.26.0. The crash started after moving to Trixie / LightDM 1.32.0.

## Suspected cause

This looks like a double-free in the XDMCP unmanaged-session timeout path in 1.32.0.
In 1.32.0, src/xdmcp-server.c uses SessionData with:

 - g_autoptr(SessionData) data = user_data;
 - g_hash_table_remove(priv->sessions, ...)

and the hash table destroy callback also frees SessionData (session_data_free), which appears consistent with:

 - double free or corruption (fasttop)
 - Source ID ... was not found when attempting to remove it

Bullseye's 1.26.0 used different session lifetime handling and did not crash in this scenario.

## Trigger in our case

[...]

PR #467, opened 2026-08-20
xdmcp-server: fix double-free on unmanaged session timeout

### LINKED ISSUE #330: XDMCP with wrong address crashes lightdm
On Rocky 9.3 with lightdm 1.30 I'm observing a crash of the lightdm master process when an XDMCP Request with an IP address is to be serviced, which cannot be connected to. After about two minutes (>= 126 seconds?) the lightdm master process crashes with these lines in the log:

```
[+3218.04s] DEBUG: Timing out unmanaged session 15759
[+3218.04s] CRITICAL: g_object_unref: assertion 'G_IS_OBJECT (object)' failed
[+3218.04s] CRITICAL: Source ID 657 was not found when attempting to remove it
```
All active sessions are terminated, too, since all children of the master process are following into Hades.

See the attached minimum C program which exercises this bug by sending the two required XDMCP packets of annihilation.

[lightdmkill.c.txt](https://github.com/canonical/lightdm/files/13618850/lightdmkill.c.txt)

ousamabenyounes:
## What

Fix a double-free crash in the XDMCP server when an **unmanaged session times out**.

Fixes #330
Also resolves the duplicate report #435.

## Why

`priv->sessions` is created with `session_data_free` as its value-destroy function:

```c
priv->sessions = g_hash_table_new_full (g_direct_hash, g_direct_equal, NULL,
                                         (GDestroyNotify) session_data_free);
```

`session_timeout_cb()` however *also* took ownership of the same `SessionData`
via `g_autoptr(SessionData)` and then called `g_hash_table_remove()`. Removing
the entry runs `session_data_free()` once; the `g_autoptr` cleanup at scope exit
runs it a **second** time — unref-ing the already-finalized `XDMCPSession` and
freeing the struct twice. `session_data_free()` additionally called
`g_source_remove()` on the timeout source that returning `G_SOURCE_REMOVE` was
already tearing down, producing the `Source ID … was not found` warning. This
matches the reported journal output:

```
g_object_unref: assertion 'G_IS_OBJECT (object)' failed
double free or corruption (fasttop)
Source ID 30 was not found when attempting to remove it
lightdm.service: Main process exited, code=killed, status=6/ABRT
```

The old code also re-read `xdmcp_session_get_id (data->session)` *after*
`g_hash_table_remove()` had freed `data` — a use-after-free.

## Fix

Make the hash table the sole owner of `SessionData`:

- drop the `g_autoptr` (and the now-unused `G_DEFINE_AUTOPTR_CLEANUP_FUNC`);
- clear `data->timeout_source` before removal so the source that
  `G_SOURCE_REMOVE` is already removing is not removed a second time;
- cache the session id before the entry is freed.

Sessions removed while their timeout is still pending (normal teardown) still
cancel their source, because `session_data_free()` keeps its
`if (timeout_source != 0) g_source_remove(...)` guard.

## Verification (AddressSanitizer RED → GREEN)

The XDMCP integration tests don't wait out the 126 s unmanaged-session timeout,
so the crash is proven with a small harness that `#include`s `xdmcp-server.c`
and drives the exact static `session_timeout_cb()` that GLib fires when the
timeout elapses (built with `-fsanitize=address`).

**Before the fix (`main`):**
```
==959==ERROR: AddressSanitizer: heap-use-after-free ... READ of size 8
SUMMARY: AddressSanitizer: heap-use-after-free src/xdmcp-server.c:157 in session_data_free
==959==ABORTING
```

**After the fix:**
```
OK: no double free
GREEN_EXIT=0
```

Full `./autogen.sh && make && make check DEBUG=true` passes on `ubuntu:rolling`
(same as CI).


jpeisach:
Hi, so I do intend on creating contributing guidelines for LightDM, and in the future, I may not accept PR's that are written by an LLM (unless it is explicitly stated and information that is included is said so below, like it is in this patch sent to LKML: https://lore.kernel.org/all/20260511221931.2370053-24-sashal@kernel.org/

Anyways, I would've rejected this, but it is a patch that I did forget about. And I actually *will* allow the LLM comments in the code this time because it's not the most obvious fix.

my only ask is that you do what I was going to do with the patch: In the commit message, at the end, add the following:
`Co-authored-by: undroidwish`

(Normally an email follows the name, but I can't seem to find one, so just leave it as is)

ousamabenyounes:
Done — `915d194f` carries `Co-authored-by: undroidwish` at the end of the commit message, name only as you said. Code is untouched: the diff against the previous head is empty.

Thanks for taking the time on this one, and for saying where you stand on LLM-assisted contributions rather than just closing it. Noted for anything I send here in future — happy to state it up front once your guidelines land.

jpeisach:
> Noted for anything I send here in future — happy to state it up front once your guidelines land.

They are being discussed in #468 - I will admit my first draft was maybe a bit too harsh?

Anyways once I figure out how to reproduce the crash in the first place and verify the fix this will be good to merge

ousamabenyounes:
> once I figure out how to reproduce the crash in the first place

Done — it reproduces in LightDM's own test suite now.

Nothing existing catches it because every XDMCP test sends `Manage` right after `Accept`. The crash needs `Manage` to *never* arrive, so `MANAGE_TIMEOUT` (126 s) fires. The runner's 4 s status timeout would normally kill that, but `#?*WAIT DURATION=` suspends it (`test-runner.c:504`). So the test is just `xdmcp-server-login.conf` minus the `SEND-MANAGE` line, plus `#?*WAIT DURATION=135`.

`make check TESTS=test-xdmcp-server-unmanaged-timeout DEBUG=true` on `ubuntu:rolling` —

main (73987eb9):

```
[+127.21s] DEBUG: Timing out unmanaged session 14018
[+127.21s] CRITICAL: g_object_unref: assertion 'G_IS_OBJECT (object)' failed
[+127.21s] CRITICAL: Source ID 32 was not found when attempting to remove it
double free or corruption (!prev)
Process 11810 terminated with signal 6
FAIL test-xdmcp-server-unmanaged-timeout
```

Same three lines as the journal in #330. With this PR (915d194f): `PASS`.

Say the word and I'll push it as a second commit — your call whether it belongs in this PR, since it adds ~130 s to `make check`.

PR #468, opened 2026-08-20
Update contributing guidelines

jpeisach:
I am open to feedback


tomekdev:
I would suggest adding a line that would state that contributor might be arbitrarily asked to explain his/her changes (in context of AI-generated code/content). If he/her cannot explain then such changes might be rejected.

I'm suggesting this because of my personal experience with vibe coders at my workplace. Some might try to "hide" that the code was vibe-coded and they have no clue what the code does, they rely only on output of AI stating that "it does ...(AI explanation)".

I know I might sound a bit like an anti-AI activist but my workplace taught me that such contributions can be valuable but also casue a huge burden for reviewers (and not rarely waste their time)

tomekdev:
In overall I really like those guidelines. There's not much that could be added to it in my personal opinion. It is short and easy to understand. Great work @JPeisach! :clap:

ayourk:
I'll admit that the https://github.com/aaif-goose/goose has a real interesting way in how they process pull requests.  I believe they use AI to review changes and suggest fixes that lead to preventing future bugs.

Fuseteam:
hmmm unsure about `All code must still be human-made`

LLMs can generate useful code snippets.

I think the focus for that should be more on responsibility, the contributor should be able to assume responsibility for the code they are submitting, LLM generated or hand written. If they can't explain the code they cannot take responsibility for it

I do get the sentiment on the comments tho, AI comments can be overly verbose— but perhaps that should be highlighted?

something similar to https://github.com/KytyPS5/KytyPS5#ai-use could also work

personally i really like the linux kernel guidelines: https://docs.kernel.org/process/coding-assistants.html

tomekdev:
@Fuseteam made a good point. Except that I saw on Phoronix that in near future kernel developers are considering deprecating `Assisted-By` tag, because it served more like an advertisement than anything useful.

ayourk:
> @Fuseteam made a good point. Except that I saw on Phoronix that in near future kernel developers are considering deprecating `Assisted-By` tag, because it served more like an advertisement than anything useful.

I agree that `Assisted-By` is not helpful.  And I can see it as an advertisement too.  We don't need that; we just need good quality code.

Fuseteam:
yeah we don't really need assisted-by per se, we can already see that with the disclosure in the related PR anyway

jpeisach:
Sorry for the delayed activity. I've been at college orientation.

The Assisted-by statement is really only helpful to the kernel developers because they have the Signed-off-by system and this adds a level of acknowledgement. Here it would just be free advertising for companies.

I don't think an AI should be running pull requests and be used to choose how to merge PRs. Also a human should review it. It's not like we are inundated with PRs that it would help, if anything it only is taking away a duty that could be assigned to a human and encouraging the economy to give up more things without people being ready for it.

I think it's best to wait and see how the Debian GR goes, and see if Ubuntu itself creates its own policy.

tomekdev:
By the way, I recently stumbled upon interesting approach for checking of someone decided to play against the rules: https://www.phoronix.com/news/NetworkManager-AI-Canary.

Funny and not harmful :smile:

Issue #371, opened 2024-08-04
Login broken with systemd 256 and homed-managed home dir


dmitrig:
Stopped being able to log in after upgrading to systemd 256. Running arch linux / xfce4 (xorg) and a systemd-homed managed home directory with default (luks) storage.

Successful auth prints:

```
Error writing X authority: Failed to open X authority /.Xauthority: Permission denied
```
and kicks me back to the password prompt.

Works fine with systemd 255 or a non-encrypted home directory. Did a bit of digging and it looks like after the upgrade `org.freedesktop.Accounts` started reporting my user's home dir as "/" while it's not activated. Tried `user-authority-in-system-dir=True` as a workaround, but that didn't work either.

Chacha75:
I've experienced the same problem in Manjaro on 3 different PCs.
Login only worked if user was logged in on another tty as mentioned above.
My solutions were to switch to GDM, or to downgrade systemd to 255.x

hmalzahn:
There is a simple workaround that works fine for me: while logged on as _your-user_, append the systemd-generated passwd-line to he system passwd database:

sudo getent passwd _your-user_ >>/etc/passwd

I guess this might have side effects, but so far I have not seen any.

hiaseilert:
I have the same problem. But home is btrfs-Subvolume (no encryption) btw.
systemd 256.x and 257.1
lightdm 1:1.32.0

Greydus:
@hiaseilert press `ctrl + alt + f2`, login as root, then run `homectl activate <username>`

PR #475, opened 2026-09-01
Fix login / subprocess execution as systemd-homed managed users

openglfreak:
Opening a PAM session may change the user's home directory and/or shell, such as with systemd-homed when the home directory has to be activated during the opening of the session.

Update the PAM environment with the new values if no module has changed the old values.

This likely fixes issue #371, although I can't reproduce that issue on the latest systemd, likely thanks to the introduction of `systemd-home-fallback-shell`.


jpeisach:
Wait, could this be why some weird stuff was going on where apps wouldn't open whenever I tried restarting lightdm via the VT?

I'm going to ask for Neal to review this, and actually @n3rdopolis also

src/session-child.c
```diff
@@ -33,6 +33,8 @@
 #include "privileges.h"
 #include "x-authority.h"
 #include "configuration.h"
+#include "accounts.h"
```
jpeisach: Optional nitpick. I know that `configuration.h` include is duplicated but could you try to make the includes list alphabetically ordered? thanks :)

(You can do your best, I won't hold this PR against you if you can't.)

src/session-child.c
```diff
@@ -550,6 +552,38 @@ session_child_run (int argc, char **argv)
     if (!home_directory) {
         home_directory = user_get_home_directory (user);
     }
+    bool pam_changed_home = strcmp(home_directory, user_get_home_directory (user)) != 0;
```
jpeisach, line 554: hm.. shouldn't the above:

```c
    /* try to get HOME from PAM since it might have been changed */
    const gchar *home_directory = pam_getenv (pam_handle, "HOME");
    if (!home_directory) {
        home_directory = user_get_home_directory (user);
    }
```

have tried to catch this?

src/session-child.c
```diff
@@ -550,6 +552,38 @@ session_child_run (int argc, char **argv)
...
+
+    const gchar *shell = pam_getenv (pam_handle, "SHELL");
+    if (!shell) {
+        shell = user_get_shell (user);
+    }
+    bool pam_changed_shell = strcmp(shell, user_get_shell (user)) != 0;
+
+    if (!pam_changed_home || !pam_changed_shell) {
```
jpeisach, line 566: does this code block even run? if either the home or the shell changed this wouldn't execute from what I can tell (or at the very least, the condition is confusing at first)

jpeisach (commented):
hopefully I can reproduce the issue in the first place

src/session-child.c
openglfreak: No, `pam_getenv` may return a different string than `user_get_home_directory`, that's what's tested for here. The PAM environment is only updated if it hasn't been changed by any PAM module yet.

src/session-child.c
openglfreak: If *either* home or shell haven't changed then this block will run, the condition can be read like `(!(pam_changed_home && pam_changed_shell))`, i.e. skip the code block if both environment variables have changed.

Edit: I got confused myself

openglfreak: Added comments to hopefully make things clearer.

src/session-child.c
openglfreak: Sorted the quote-includes in a separate commit added to the start.

openglfreak:
Hmm, seems like this doesn't fix my issue after all, systemd just kept my home partition activated after logging out (which sounds like a bug in itself) and that made my problems go away after logging back in...

openglfreak:
I have forced AccountsService to uncache my user using `org.freedesktop.Accounts.UncacheUser` and now everything works again (with this PR), I am incredibly confused, in any case this is shouldn't be a negative change.

jpeisach:
Hm. I am confused in the sense that I am still learning about all of this. Hopefully Neal can review and I'll do my homework on PAM

Conan-Kudo:
I'll try to take a look later this week. I've got a bunch of stuff to proc through first.

PR #476, opened 2026-09-02
Keep a file descriptor open for reserved VTs

Mrmaxmeier:
Hi,

I recently looked into odd behaviour with LightDM on NixOS:
https://github.com/NixOS/nixpkgs/issues/422331#issuecomment-5510428213
NixOS sets `minimum-vt = 1` and suffers from a race where a VT is used both by the desktop session and getty. In this situation, getty then receives desktop session keystrokes and resets the keyboard out of K_OFF. As a symptom, Alt+Left/Right then switch VTs from inside of the desktop session.

It seems like this can be mitigated by holding onto VT fds inside of LightDM: logind skips VTs that are "in use" / where fds are held by any process.

Note: This change in LightDM is LLM-suggested but looks fine to me. I'm running the patched version and while the underlying mechanism to avoid collisions doesn't seem fully sound to me, it seems to mitigate the racy behaviour.

Does this change make sense or is NixOS holding display managers wrong in this case?

Thanks! :slightly_smiling_face:


jpeisach:
> Does this change make sense or is NixOS holding display managers wrong in this case?

/shrug

I'll come back to this.. eventually! :D

Thanks for disclosing your LLM usage

PR #478, opened 2026-09-03
perf: optimize user-list, xauth, and xdmcp paths

GermanAizek:
@jpeisach,

Using perf benchmarks from this PR (https://github.com/ubuntu/lightdm/pull/477), we were able to carry out useful algorithmic optimization, namely:

User List
   - Replace O(N^2) per-entry g_list_insert_sorted() during passwd loading with O(1) prepend and a single O(N log N) g_list_sort() merge sort
   - Replace O(N^2) nested list traversals for existing and removed users with O(1) hash tables

Xauthority
   - Buffer all serialized records into a GByteArray in memory before writing to disk in a single atomic write() syscall, replacing up to 350 separate syscalls per update
   - Replace byte-by-byte loops in read_data() with block memcpy()
   - Replace O(N^2) g_list_append() with O(N) prepend + reverse

XDMCP
   - Replace byte-by-byte loops in read_data(), read_string(), write_data(), and write_string() with single-check block memcpy() operations
   
I post visual graphs my benchmarks on NUMA server motherboard 2x Xeon E5-2699v3 (Haswell 2014 year - 36 core 72 threads)

Two commits were compared: 
- integrate-benchmarks branch https://github.com/ubuntu/lightdm/pull/477/commits/4bf60bab405f08ecff846710f308bd90f06049d0 (w/o optimizations)
- optimize branch https://github.com/ubuntu/lightdm/commit/1dd9c812c6bf338cb4aff3cd070ef46f1caf8800

Attention on lookup_by_name from user list, change math algorithm speedup in 371x times !!!

image

image

image

image

### DIFF
--- a/common/user-list.c
+++ b/common/user-list.c
@@ -347,6 +347,13 @@ load_passwd_file (CommonUserList *user_list, gboolean emit_add_signal)
 
     setpwent ();
 
+    GHashTable *existing_users_map = g_hash_table_new (g_str_hash, g_str_equal);
+    for (GList *link = priv->users; link; link = link->next)
+    {
+        CommonUser *u = link->data;
+        g_hash_table_insert (existing_users_map, (gpointer) common_user_get_name (u), u);
+    }
+
     GList *users = NULL, *new_users = NULL, *changed_users = NULL;
     while (TRUE)
     {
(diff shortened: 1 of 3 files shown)


jpeisach:
So the thing with the user lists.. I imagine most systems don't more than like, eight users at most? So I'm not sure how much this helps. Unless there are people using LightDM on systems with *tons* of users (in which this would be good)

I'm definitely interested in Xauthority and XDMCP optimizations.

I haven't decided what balance between space and time complexity we should have..

GermanAizek:
> Unless there are people using LightDM on systems with _tons_ of users

This is easy to calculate as a proportion.

371x - 1000 users lookup
37.1x - 100 users
3.7x - 10 users performance up

It’s actually more logical to use a hashtable.

jpeisach:
By that calculation.. for a one user system that's a slowdown (0.37)

